PR #126, opened 2018-09-10
Get started

amn41:


data/core/get_started.md
```diff
@@ -0,0 +1,175 @@
...
+
+
+## Generated Story 4741357980983644475
+* greet
+    - utter_greet
+    - utter_inform_privacypolicy
+* mood_confirm
+    - utter_awesome
+    - utter_ask_goal
+* how_to_get_started{"product": "core"}
```
akelad: we need a story for "product":"nlu" too

data/core/get_started.md
```diff
@@ -0,0 +1,175 @@
...
+  - utter_quickstart_nlu_only
+* thank
+  - utter_noworries
+
+
+## story_004
+* greet
+    - utter_greet
+    - utter_inform_privacypolicy
+* mood_confirm
```
akelad: this privacy policy flow needs to be changed everywhere
From:
```
* greet
    - utter_greet
    - utter_inform_privacypolicy
* mood_confirm
    - utter_awesome
    - utter_ask_goal
```
to
```
* greet
    - utter_greet
    - utter_inform_privacypolicy
    - utter_ask_goal
```

data/core/get_started.md
akelad: Also a stories where the user says `* how_to_get_started` and then afterwards says he wants to know more about core or NLU.

data/nlu/nlu.md
```diff
@@ -1256,3 +1255,72 @@
...
+- what components of Rasa are open source
+- how to integrate rasa chatbot with my website
+- how works rasa
+- with exceptions
+- which language is rasa programmed in
+- what technologies did u use to create more mature chatbot?
+- which technology is used to create you
+
+
+## intent:switch
```
akelad: i think we also need to add some examples of utterances where the user just says "core"/"dialogflow" etc

akelad: but all this training data needs to be added to the platform anyways, not here

data/core/get_started.md
```diff
@@ -0,0 +1,175 @@
...
+  - utter_quickstart
+  - utter_ask_migration
+* switch{"current_api": "luis"}
+  - utter_switch_luis
+
+## story_003
+* how_to_get_started
+  - utter_quickstart
+  - utter_ask_migration
+* switch
```
akelad: this story doesn't make much sense 😄

data/core/get_started.md
```diff
@@ -0,0 +1,175 @@
...
+    - utter_greet
+    - utter_inform_privacypolicy
+* mood_confirm
+    - utter_awesome
+    - utter_ask_goal
+* how_to_get_started
+  - utter_quickstart
+  - utter_ask_migration
+* switch
+  - utter_quickstart_nlu_only
```
akelad: same here, `utter_quickstart_nlu_only` shouldn't come after the `switch` intent

domain.yml
```diff
@@ -169,6 +195,8 @@ templates:
     - text: "I can't sign you up if you don't provide your email 😉"
   utter_ask_isbot:
     - text: "Yep I'm a bot!"
+  utter_cant_answer_tech_question:
+    - text: "I can't answer that"
```
akelad: I think we should change this to "I can't answer that, please take a look at our docs to find your answer: (link)", but we can change that in the platform rather than here

data/core/get_started.md
```diff
@@ -0,0 +1,175 @@
...
+## tech question again
+* greet
+    - utter_greet
+    - utter_inform_privacypolicy
+* mood_confirm
+    - utter_awesome
+    - utter_ask_goal
+* technical_question
+  - utter_cant_answer_tech_question
+  - utter_docu
```
akelad: the `utter_docu` utterance should be removed from the flows with `* technical_question`

data/core/get_started.md
amn41: this covers the case where someone says something like "I am switching from x" where x is some random product. Doesn't it make sense to just push them to the nlu quickstart in that case?

data/core/get_started.md
amn41: what would you suggest?

data/core/get_started.md
akelad: hmmm, i mean they could also say something like "I'm switching", in which case we'd say "From what service?", and then maybe after that send the NLU quickstart link?

amn41: I'd prefer to keep it simple for now and leave out the extra qeustion. maybe just sending both luis and dialogflow links?

PR #137, opened 2018-10-17
Update Dockerfile for Python 3

claverru:
Make image FROM Python:3.6-slim
- get Makefile working
- coherent README


akelad:
hm yes good point, thanks for that. Not entirely sure why this wasn't python3 before given I developed it in python 3 :D

PR #222, opened 2018-11-29
extended chitchat with new intents

EPedrotti:


domain.yml
```diff
@@ -28,6 +28,15 @@ intents:
   - ask_whatspossible
   - ask_whatisrasa
   - ask_isbot
+  - ask_howold
```
akelad: please add corresponding `utter_` acctions to actions list

data/core/chitchat.md
```diff
@@ -1,5 +1,5 @@
 ## chitchat
-* ask_weather OR ask_builder OR ask_whatspossible OR ask_howdoing OR ask_whoisit OR ask_whatisrasa OR ask_isbot
```
akelad: why did you get rid of ask_whatspossible?

akelad: in loads of places 😠

PR #236, opened 2018-12-14
Step3 improvements

wochinge:
**Proposed Changes**:
* add stories for the case if users provide their package manager early (e.g. "I want to install Rasa with pip"
* add a faq question for the case users ask which python version they should use


wochinge:
Thanks, good catch! Pushed the change :-)

PR #253, opened 2018-12-18
Step4 Implementation

wochinge:
Steps:
- [x] ready to review

### DIFF
--- a/data/core/step4.md
+++ b/data/core/step4.md
@@ -3,7 +3,7 @@
     - action_greet_user
     - utter_describe_step4
     
-## Happy path: Prompt -> Get event locations -> get date for location 
+## Happy path: Prompt -> Get event locations -> get next event 
 * get_started_step4
     - action_greet_user
     - utter_describe_step4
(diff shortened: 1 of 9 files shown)


data/core/step4.md
```diff
@@ -0,0 +1,60 @@
...
+## Prompt -> How can I contribute
+* get_started_step4
+    - action_greet_user
+    - utter_describe_step4
+* ask_how_contribute
+    - utter_possibilities_to_contribute
+    
+## How can I contribute
+* ask_how_contribute
+    - utter_possibilities_to_contribute
```
akelad: What about the "I want to ask a question on the forum" flow?

demo/actions.py
```diff
@@ -485,3 +487,86 @@ def submit(self,
...
+    def __init__(self):
+        self.last_event_update = None
+        self.events = None
+        self.events = self._get_events()
+
+    def name(self) -> Text:
+        return "action_get_community_events"
+
+    def _get_events(self) -> List['CommunityEvent']:
+        if self.events is None or self._are_events_expired():
```
akelad: hmm, do we need any of these checks? because every time you call the action, self.events will be None :P

demo/actions.py
```diff
@@ -485,3 +487,86 @@ def submit(self,
...
+            from demo.community_events import get_community_events
+            logger.debug("Getting events from website.")
+            self.last_event_update = datetime.now()
+            self.events = get_community_events()
+
+        return self.events
```
akelad: I think something that would be useful in the `get_events()` function (or `get_community_events()` I guess), is a check that the events are actually in the future. on our website we have a lot of events that are in the past

demo/actions.py
wochinge: Are the action objects re-created between the requests? When I check the logs of the action endpoint it actually just calls `_get_events` when the server is started or the events are expired

akelad: Hm yes actually good point, should be fine

PR #262, opened 2019-01-10
add transition steps

akelad:


demo/actions.py
```diff
@@ -426,8 +426,13 @@ def run(self,
             ) -> List['Event']:
 
         intent_ranking = tracker.latest_message.get('intent_ranking', [])
+        diff_intent_confidence = intent_ranking[0].get("confidence") - intent_ranking[1].get("confidence")
+        if len(intent_ranking) > 1 and diff_intent_confidence < 0.2:
```
wochinge: the check `len(intent_ranking) > 1` is a bit late since you do `intent_ranking[1].get("confidence")` in the line before

demo/actions.py
```diff
@@ -583,3 +588,21 @@ def _utter_next_event(self,
...
+class ActionNextStep(Action):
+
+    def name(self):
+        return "action_next_step"
+
+    def run(self, dispatcher, tracker, domain):
+        step = tracker.get_slot('step')
+
+        button = [{'title': 'Next step',
+                   'payload': '/get_started_step{}'.format(step)})
```
wochinge:
```suggestion
                   'payload': '/get_started_step{}'.format(step)}]
```

domain.yml
```diff
@@ -623,6 +626,7 @@ actions:
   - utter_possibilities_to_contribute
   - utter_recommend_forum
   - utter_link_to_forum
+  - utter_direct_step3
```
wochinge: Let's add a template text

demo/actions.py
akelad: good point 🤦‍♀️

demo/actions.py
akelad: good catch thank you

Issue #274, opened 2019-01-28
Add tutorial intent


akelad:
"Are there tutorials for rasa_core?"

PR #282, opened 2019-02-18
added ask tutorial intent

imLew:
fixes https://github.com/RasaHQ/rasa-demo/issues/274

Basically whenver a user enters something like "Can you show me some tutorials?" or "Do you have tutorials for Rasa?" Sara will reply with "Yes, here are some tutorials I wrote, but also check out these great community tutorials"

Something to think about would be using synonyms for tutorial, I think it's unlikely but some people might use words like "walkthrough".

Also I though it would be funny if Sara said she wrote the tutorials written by people at Rasa

### DIFF
--- a/domain.yml
+++ b/domain.yml
@@ -429,7 +429,7 @@ templates:
   utter_get_started_step3:
     - text: "I can help you to install Rasa Stack."
   utter_ask_faq_tutorials:
-    - text: "Yes, [here](https://blog.rasa.com/tag/tutorials/) are some tutorial I wrote, but you can also check out some great tutorials written by the Rasa community [here](https://docs.google.com/spreadsheets/d/1lzvSRsTr0mQdJT_usvo1jMDzHjzvjVbiUkHpiGIKUgA/edit#gid=1717872446)"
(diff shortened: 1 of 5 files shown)


domain.yml
```diff
@@ -427,6 +428,8 @@ templates:
     - text: "I can help you solving commonon problems with Rasa 🙂"
   utter_get_started_step3:
     - text: "I can help you to install Rasa Stack."
+  utter_ask_faq_tutorials:
+    - text: "Yes, [here](https://blog.rasa.com/tag/tutorials/) are some tutorial I wrote, but you can also check out some great tutorials written by the Rasa community [here](https://docs.google.com/spreadsheets/d/1lzvSRsTr0mQdJT_usvo1jMDzHjzvjVbiUkHpiGIKUgA/edit#gid=1717872446)"
```
wochinge: Should we maybe give a list of tutorial links until the docs page is finished?

imLew: There are too many.
I want to try and do the docs this week (though I'm not sure how much time that takes).
But if we want to deploy this before the docs are finished I suggest we just leave out the community tutorials for now and just add them once the docs are done.

imLew: I just made that change

akelad: yes that sounds good @imLew

PR #285, opened 2019-03-03
Update nlu.md

mrjazz:
Ukraine country added


akelad:
thanks for the addition!

PR #309, opened 2019-04-15
update domain file so that all actions work locally

erohmensing:
- adds utterances that were called in the actions file but were not available as templates
  - completes fix of changing `howwereyoubuilt` to `ask_howbuilt` everywhere


akelad:
omg i can't believe that was called `howwereyoubuilt` 🤦‍♀️

domain.yml
```diff
@@ -537,6 +547,8 @@ templates:
 actions:
   - utter_ask_faq_tutorials
   - utter_greet
+  - utter_greet_name
```
akelad: we'll leave this here, but for educational purposes: you don't actually need to add all templates if they're only called from an action and not used in the stories :D

erohmensing: good to know! You just mean in the list actions, right? That makes sense.

akelad: exactly

PR #325, opened 2019-06-11
Remaining updates to Rasa 1.0

tabergma:
- update requirements
- remove not needed imports
- update Makefile
- use `rasa_sdk` instead of `rasa_core_sdk`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,36 +3,22 @@
 TEST_PATH=./
 
 help:
-	@echo "    train-nlu"
-	@echo "        Train the natural language understanding using Rasa NLU."
-	@echo "    train-core"
-	@echo "        Train a dialogue model using Rasa core."
+	@echo "    train"
+	@echo "        Train a Rasa model."
 	@echo "    run-cmdline"
 	@echo "        Starts the bot on the command line"
 	@echo "    visualize"
 	@echo "        Saves the story graphs into a file"
 
 run-actions:
-	python3 -m rasa_core_sdk.endpoint --actions demo.actions
+	rasa run actions --actions demo.actions
 
-train-nlu:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -3,36 +3,28 @@
...
 visualize:
-	python3 -m rasa_core.visualize -s data/core/ -d domain.yml -o story_graph.png
-
-train-online:
-	python -m rasa_core.train -u models/nlu/current/ --online --core models/dialogue/
+	rasa visualize --stories data/core/ --domain domain.yml --out story_graph.png
 
 evaluate-core:
-	python -m rasa_core.test --core models/dialogue -s data/core/ --fail_on_prediction_errors
+	rasa test core --model models/dialogue --stories data/core/ --fail-on-prediction-errors --quiet
```
akelad: why quiet?

Makefile
```diff
@@ -3,36 +3,28 @@
...
-	python3 -m rasa_nlu.train -c nlu_tensorflow.yml --fixed_model_name current --data data/nlu/ -o models --project nlu --verbose
-
-train-core:
-	python3 -m rasa_core.train -d domain.yml -s data/core -c policy.yml --debug -o models/dialogue
+train:
+	rasa train
 
 train-memo:
-	python -m rasa_core.train -d domain.yml -s data/core -c augmentedmemo-only.yml -o models/dialogue --augmentation 0
+	rasa train core --domain domain.yml --stories data/core --config augmentedmemo-only.yml --out models/dialogue --augmentation 0 --quiet
```
akelad: why quiet?

Makefile
tabergma: travis failed and complaint that there is too much output

Makefile
tabergma: same

Makefile
akelad: hahaha that's a new one, it used to fail when there wasn't any output for a while

PR #330, opened 2019-07-12
Add Rasa X to Getting Started flow

msamogh:


data/core/chitchat.md
```diff
@@ -425,9 +441,9 @@
...
 * deny
-    - utter_explain_stack
-    - utter_stack_details
-    - utter_explain_nlucore
+    - utter_explain_rasa_components
+    - utter_rasa_components_details
+    - utter_ask_explain_nlucorex
 * how_to_get_started{"product": "core"}
     - utter_explain_core
     - utter_also_explain_nlu
```
erohmensing: So to be clear, if they say they want to get started with everything, we explain nlu, core, and x, but if they say they want to get started with nlu we explain nlu and core and vice versa (but not x) -- is this intended?

erohmensing: just noticed below that when you explain x, you also explain NLU and Core, so this is the only case where everything is not explained

data/core/closetheloop.md
```diff
@@ -1,921 +0,0 @@
...
-* ask_whoisit
-    - action_chitchat
-* enter_data
-    - utter_not_sure
-    - utter_possibilities
-* ask_faq_languages
-    - action_faqs
-* enter_data
-    - utter_not_sure
-    - utter_possibilities
```
erohmensing: still confused as to why you've deleted this whole file

data/core/faqs.md
```diff
@@ -146,410 +141,6 @@
...
-    - utter_explain_stack
-    - utter_stack_details
-    - utter_explain_nlucore
-* affirm OR how_to_get_started{"product":"stack"}
-    - utter_explain_nlu
-    - utter_explain_core
-    - utter_direct_to_step2
-* ask_faq_platform OR ask_faq_languages OR ask_faq_tutorialcore OR ask_faq_tutorialnlu OR ask_faq_opensource OR ask_faq_voice OR ask_faq_slots OR ask_faq_channels OR ask_faq_differencecorenlu OR ask_faq_python_version OR ask_faq_community_size OR ask_faq_what_is_forum OR ask_faq_tutorials
-    - action_faqs
-    - utter_direct_to_step2
```
erohmensing: as well as most of this file

data/core/faqs.md
```diff
@@ -748,28 +339,6 @@
...
-    - utter_explain_stack
-    - utter_stack_details
-    - utter_explain_nlucore
-* affirm OR how_to_get_started{"product":"stack"}
-    - utter_explain_nlu
-    - utter_explain_core
-    - utter_direct_to_step2
-
 ## FAQ - tell more about platform
 * ask_faq_platform
```
erohmensing: platform?

data/core/step1_get_started.md
```diff
@@ -25,11 +25,36 @@
...
+    - utter_rasa_components_details
+    - utter_ask_explain_nlucorex
+* affirm OR how_to_get_started{"product":"all"}
+    - utter_explain_nlu
+    - utter_explain_core
+    - utter_explain_x
+    - utter_direct_to_step2
 
-## new to rasa + not new to chatbots + not migrating
+## new to rasa + not new to chatbots + migrating from
```
erohmensing: this might be one of those weird git diff things that happens with stories, but this user isn't migrating

data/core/step1_get_started.md
```diff
@@ -149,12 +175,13 @@
...
     - utter_built_bot_before
 * deny
-    - utter_explain_stack
-    - utter_stack_details
-    - utter_explain_nlucore
-* affirm OR how_to_get_started{"product":"stack"} OR explain
+    - utter_explain_rasa_components
+    - utter_rasa_components_details
+    - utter_ask_explain_nlucorex
+* affirm OR how_to_get_started{"product":"all"} OR explain
```
erohmensing: can we add OR explain to the others? looking at the NLU data it seems to make sense here

data/core/step2_rasa_init.md
```diff
@@ -0,0 +1,15 @@
...
+* affirm
+    - utter_run_rasa_init
+    - utter_direct_to_step3
+
+## Get starter from website + deny
+* get_started_step2
+	- action_greet_user
+	- slot{"step": "2"}
+* deny
+	- utter_anything_else
```
erohmensing: weird formatting in this one

data/nlu/faq.md
```diff
@@ -0,0 +1,284 @@
...
+- which messaging channels are supported by rasa?
+- rasa supports which messaging channels?
+- which messaging channels can I use with rasa?
+- which messaging channels can be used with rasa?
+- which messaging channels are compatible with rasa?
+- what are the messaging channels that can be used with rasa?
+- rasa can use which different messaging channels?
+- WHAT MESSAGING PORTALS DOES RASA SUPPORT?
+- DOES RASA SUPPORT MESSENGER?
+- DOES RASA SUPPORT THE WHATS APP?
```
erohmensing: omg ahhaha these are so aggressive

domain.yml
```diff
@@ -1,753 +1,864 @@
...
+  utter_search_bar:
+  - text: Try using the search bar above for now.
+  utter_source_code:
+  - text: You can find my source code [here](https://github.com/RasaHQ/rasa-demo/).
+  utter_spacy:
+  - text: 'SpaCy has excellent pre-trained named-entity recognizers for this -- check
+      out this [interactive demo](https://demos.explosion.ai/displacy-ent/). I would
+      suggest using the SpacyEntityExtractor component in your pipeline.
+
+      '
```
erohmensing: I think we can move some of these strange trailing quotes up to the end of the statement (e.g. right after pipeline) -- not sure though

data/core/closetheloop.md
msamogh: It's been renamed to `interactive.md`.

erohmensing: maybe i'm blind but i don't see that file anywhere, did it get lost in the PR port?

data/core/faqs.md
msamogh: We were planning to keep the old platform FAQs as they did no harm (and didn't turn up with nothing, for someone who is used to using that term)

data/core/closetheloop.md
msamogh: My bad. I think it got lost there. I've added it now.

data/core/chitchat.md
msamogh: Core and NLU seem like two siblings that can't live without each other. So if they're asking about one of Core/NLU, we might be serving to unnecessarily extending the conversation by offering to about Rasa X.

data/core/faqs.md
erohmensing: got it

PR #357, opened 2019-11-28
revamp dockerfile, upgrade to 1.5.1

erohmensing:
* get docker file from rasa-sdk one 
* ignore everything but `demo` folder in setup.py
* remove rasa from setup.py
* separate requirements of the app versus the whole bot
* uprade to 1.5.1


rgstephens:
I ran Sara locally under the updated Dockerfile, tried out a few intents that required calls to the action agent, including `ActionDefaultAskAffirmation`, and it worked. This is my first review, let me know if I should run any other tests.

erohmensing:
> I ran Sara locally under the updated Dockerfile

Awesome thanks! To be clear, it was only running the action server right?

I say we should merge this first so that I can update our instance to 0.23.1, but I'm also intruigued by your latest issue. 

In addition to reviewing for functionality, if you have any codestyle/quality suggestions, feel free to comment on those as well (esp in the dockerfile, as i'm not the most experienced there)

rgstephens:
I decided to confirm that a 1.5.1 rasa shell could talk to your updated action container and it can.

PR #359, opened 2019-11-29
Run local docker

rgstephens:
Adds a `docker-compose-local.yml` to run Duckling and the action agent as containers and documents the commands to run the training and shell under docker. Also the associated `config_local.yml` and `endpoints_local.yml`.


erohmensing:
@rgstephens i'm going to assume you accidentally tagged emma instead of me 😄

rgstephens:
Yes, thank you

rgstephens:
Ella and I spoke, will cancel for now, update issue with new thoughts

Issue #351, opened 2019-11-20
ActionDefaultAskAffirmation not working properly


akelad:
In this interaction, it should be sending me this message instead: https://github.com/RasaHQ/rasa-demo/blob/master/demo/intent_description_mapping.csv#L49
![image (6)](https://user-images.githubusercontent.com/32450038/69274335-2916e900-0ba8-11ea-82de-1e1b38963e85.png)

ArjaanBuijk:
Now it works. Two examples:

![image](https://user-images.githubusercontent.com/16159200/70352744-d5f49580-1839-11ea-8f27-56c9a181e583.png)

![image](https://user-images.githubusercontent.com/16159200/70352774-e60c7500-1839-11ea-9345-0fcb0ee111f7.png)

PR #413, opened 2019-12-06
ActionDefaultAskAffirmation - Fix indexing into pandas data-frame

ArjaanBuijk:
Fix for Issue #351 

The multi-indexing into the pandas Dataframe did not work.

Now it works OK. Here two examples:

image

image


ArjaanBuijk:
Going to do it slightly different

PR #415, opened 2019-12-07
Use parentheses around the boolean expressions

ArjaanBuijk:
Issue #351 

The parentheses are mandatory since & has a higher operator precedence than ==.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -6,7 +6,7 @@ pipeline:
   OOV_token: oov
   token_pattern: (?u)\b\w+\b
 - name: CountVectorsFeaturizer
-  analyzer: "char_wb"
+  analyzer: char_wb
   min_ngram: 1
   max_ngram: 4
 - name: EmbeddingIntentClassifier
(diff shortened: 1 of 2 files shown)


akelad:
@ArjaanBuijk seems like you didn't add the file that's actually relevant to this PR?

ArjaanBuijk:
@akelad , I pushed in the correct file now (actions.py)
Note I also pulled the actions.py through `black`, so one more piece of code was changed, but that is just a reformat.

config.yml
```diff
@@ -19,12 +19,13 @@ pipeline:
...
 - name: EntitySynonymMapper
-
 policies:
 - name: EmbeddingPolicy
   max_history: 10
   epochs: 20
-  batch_size: [32,64]
+  batch_size:
+  - 32
+  - 64
```
akelad: what happened here?

erohmensing: looks like a rasa x dump, should be fine

PR #416, opened 2019-12-09
Run Sara w/Rasa X & Mr. Bot locally

rgstephens:
This PR adds a `docker-compose` setup that will bring up Sara with Rasa X and Mr. Bot under the local setup. It adds a `docker-compose-local.yml` and the `docker-local`.

The following commands build the images and start the containers after which you can browse to `http://localhost:5002` and `http://localhost:8080`

```
docker-compose -f docker-compose-local.yml build --no-cache --build-arg RASA_X_VERSION=0.23.3
docker-compose -f docker-compose-local.yml up -d
docker-compose -f docker-compose-local.yml logs | grep password  # to get the Rasa X password
```

You need to train the model under Rasa X, `rasa train` or with docker:

```
docker run -v $(pwd):/app rasa/rasa:1.5.1-full train --augmentation 0 --config docker-compose/config.yml
```


akelad:
I'm not sure we need to add all this stuff: 
- Rasa X should be deployed on a server
- the Mr bot widget can be pulled without building it as a docker container

PR #417, opened 2019-12-09
Fix config

JEM-Mosig:
It looks like the `AugmentedMemoizationPolicy` and the `max_history` setting lines are in the wrong order. I just fixed this.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -25,8 +25,8 @@ policies:
   max_history: 10
   epochs: 20
   batch_size: [32,64]
-- max_history: 6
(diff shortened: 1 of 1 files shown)


erohmensing:
Is this really an issue? It's just that it dumps in alphabetical order. Same with `core_threshold` coming before `name` in the fallback policy

JEM-Mosig:
@erohmensing Oh that's how you read it! I was wondering why it works at all. Ok, never mind then.

erohmensing:
Sorry, thought that was clarified in Akela's post

PR #418, opened 2019-12-11
Update Action_greet_user

akelad:
we want to make sure that if the intent is `enter_data` that we also reply to the user saying "Hi akela"


erohmensing:
do we have a story for this case already?

akelad:
as far as i know we do, i haven't checked, but i definitely wrote one at some point in life. And I think when I was testing it it triggered the `action_greet_user`, but then didn't respond to me because we're not handling this case

PR #425, opened 2019-12-20
Rasa X annotations

erohmensing:
very first rasa x annotations! wanted to get the initial dump out of the way (looks like it wanted to reorder synonyms) and i added some technical questions as well

### DIFF
--- a/data/nlu/nlu.md
+++ b/data/nlu/nlu.md
@@ -347,6 +347,15 @@
 - I need to know about Rasa X
 - please tell me more about rasa x
 - elaborate on rasa x
+- Can you tell me about rasa x?
+- I want information about rasa x
+- Rasa X
+- Rasa X features
+- Tell me about rasa x
+- can someone help me with infos about rasa x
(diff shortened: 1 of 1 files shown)


data/nlu/nlu.md
```diff
@@ -4724,6 +4722,27 @@
 - which technology is used to create you
 - yes what if i have to code open end responses into some categories
 - your specifications
+- best policies to be used
+- Deploy to a Server
+- Hello, Sara. How can I configure etnity extraction for russian lnguage?
+- Is there API for any other programming languages?
+- can I use Rasa with my Raspberry Pi
+- how to setup rasax on slack
```
ArjaanBuijk: `rasax` could be added as a new synonym for `x`

erohmensing:
good point! I think in the future we actually need to fix these (rasa and such should be annotated as `product` everywhere, and then we can just ignore the entities for featurization in this intent), but i will add as a synonym as well.

PR #430, opened 2019-12-30
Annotated several conversations

ArjaanBuijk:
I did all the annotations with Rasa X of website-demo, except the addition of synonym `custom actions`
Note that Rasa X re-arranged some synonyms compared to master, without any content change.

### DIFF
--- a/data/nlu/nlu.md
+++ b/data/nlu/nlu.md
@@ -4859,6 +4859,11 @@
 - Core
 - CORE
 
+## synonym:custom actions
+- actions
+- Custom actions
+- Actions
+
 ## synonym:date
 - dates
 
--- a/data/nlu/nlu.md
+++ b/data/nlu/nlu.md
@@ -1991,6 +1991,7 @@
 - who talks?
 - whom i talking to
 - yes but who are you in your heart
+- may i know your name
 
 ## intent:ask_why_contribute
 - What can I bring to help your code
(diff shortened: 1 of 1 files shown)


data/nlu/nlu.md
```diff
@@ -3045,6 +3046,7 @@
 - we think 4 million INR/ year
 - we're building a conversational assistant for our employees to book meeting rooms.
 - £50k
+- botonic
```
erohmensing: is this in the context of a user giving their company? if so we should annotate it as such

data/nlu/nlu.md
```diff
@@ -4846,6 +4859,11 @@
 - Core
 - CORE
 
+## synonym:custom actions
+- actions
+- Custom actions
+- Actions
```
erohmensing: Agreed this is a good idea bc it'll improve greg's search. did you add this?

data/nlu/nlu.md
erohmensing: (same with wordpress and slack, actually)

ArjaanBuijk: Both were in the context of a user providing information related to switching from another platform. The story expects the `enter_data` intent

data/nlu/nlu.md
ArjaanBuijk: Yes, I indeed added this with Greg's search update in mind. I did not yet test with Greg's search, but plan to do so once that one is merged.

PR #433, opened 2019-12-31
Update readme with action server info

rgstephens:
This PR updates the README with info on how the action server is built and deployed as well as the keys required for the MailChimp and Algolia API's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Use `rasa train` to train a model (this will take a significant amount of memory
 if you want to train it faster, try the training command with
 `--augmentation 0`).
 
-Then, to run, first set up your action server in one terminal window:
+Then, to run, first set up your action server in one terminal window. To run the MailChimp and Alogolia docs search, you must first set the associated environment variables for those API's. These settings can be found in the `docker-compose.override.yml` file on the production server.
 ```bash
 rasa run actions --actions demo.actions
 ```
@@ -63,7 +63,7 @@ your website.
 After doing a `rasa train`, run the command:
 
 ```bash
-rasa test nlu -u test/test_data.json --model models
+rasa test nlu -u test/test_nlu.md --model models
 rasa test core --stories test/test_stories.md
 ```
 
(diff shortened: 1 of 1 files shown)


rgstephens:
The readme needs to add info on how to `docker login` from the server.  You can't just change the version of the action server image in the `docker-compose.yml`. Must login before you can pull the repo. There's a `gcr-auth.json` file that maybe used for the login. Not sure.

The travis file does:

`docker login -u _json_key -p "$(cat travis_gcloud_auth.json)" https://gcr.io`

erohmensing:
I agree that this needs to be written somewhere for us, but the README is for the general public (see `There are some custom actions that require connections to external services, specifically SubscribeNewsletterForm and SalesForm. For these to run you would need to have your own MailChimp newsletter and a Google sheet to connect to.`), not for us using the server -- i feel like this info might be confusing as they dont have access to these servers/repos

README.md
```diff
@@ -35,7 +35,7 @@ Use `rasa train` to train a model (this will take a significant amount of memory
 if you want to train it faster, try the training command with
 `--augmentation 0`).
 
-Then, to run, first set up your action server in one terminal window:
+Then, to run, first set up your action server in one terminal window. To run the MailChimp and Alogolia docs search, you must first set the associated environment variables for those API's. These settings can be found in the `docker-compose.override.yml` file on the production server.
```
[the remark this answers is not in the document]
ArjaanBuijk: And remove any reference to `docker-compose.override.yml`.

README.md
```diff
@@ -79,6 +79,12 @@ rasa test core --stories test/test_stories.md
 
 `config.yml` - training configurations for the NLU pipeline and policy ensemble
 
+## Action Server
```
ArjaanBuijk: I like this description a lot, but having a `Action Server` section by itself does not help I think. I also think that we should store our version of `docker-compose.override.yml` in this repo, off course with the actual keys removed. 

I propose that you remove this section from the current README, but then create a separate issue for an enhancement to the README to include a section `How to deploy Sara to a server using Rasa X`, and that we then add these things like including the `docker-compose.override.yml` as part of that new issue.

ArjaanBuijk: To make things clearer in the current README that we only cover installation and running locally, you could update the titles and descriptions of those sections:

## Installation locally
To install Sara locally, ...

## To run Sara locally
...

PR #454, opened 2020-01-14
Rasa X annotations - Demo'ing

btotharye:
Just demo'ing the git integration


btotharye:
@ArjaanBuijk not sure if we want to merge this in, I was demo'ing stuff last night and forgot once I create this it lost the changes so we should probably merge this in, thoughts?

ArjaanBuijk:
@btotharye ,
I checked out the changes, and nothing changed except one line that was just for testing.

I think we should just close this PR without merging.

PR #455, opened 2020-01-16
Rasa X annotations

erohmensing:
Rasa X was in a red state after I added lookup tables, so I dumped some changes.

Looks like Rasa X wants to dump the lookup files in `data/location.txt` instead of in our `data/nlu/lookups/location.txt`. We can put that in now, but might be worth asking Tobi if that's really the desired place they should go or if that should be configurable (i.e. like it finds our nlu file where it is instead of trying to dump to `data/nlu.md`

### DIFF
--- a/data/nlu/nlu.md
+++ b/data/nlu/nlu.md
@@ -730,6 +730,7 @@
 - would you call rasa open source software?
 - would you consider rasa open source software?
 - How is it opensource
+- about Rasa OPen Source
 
 ## intent:ask_faq_python_version
 - What Python version should I use?
@@ -1260,6 +1261,7 @@
 - when is your birthday?
 - when were you born?
 - when you were bon
+- how old r u ?
 
 ## intent:ask_isbot
 - Ar you a bot ?
@@ -3125,6 +3127,7 @@
 - wordpress
 - slack
 - e commerce bot
+- cicciopasticcio
 
 ## intent:explain
 - can you elaborate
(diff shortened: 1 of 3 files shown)


data/nlu/nlu.md
```diff
@@ -4319,6 +4322,7 @@
 - you will know it from the single red rose it carries between its teeth
 - you're a woman
 - ı am learning python
+- cosa posso fare?
```
erohmensing: should go in the new out of scope non-english after fixing merge conflicts

data/nlu/nlu.md
```diff
@@ -3125,6 +3127,7 @@
 - wordpress
 - slack
 - e commerce bot
+- cicciopasticcio
```
erohmensing: should some of these be annotated with [company]? maybe that is already in master as well, not sure.

data/nlu/nlu.md
```diff
@@ -4824,6 +4829,7 @@
 - how to add dropdowns?
 - replace default nlu with custom component
 - can i programm a vocal assistant
+- connect rasa to my own website
```
erohmensing: this should go in ask_faq_channels

data/nlu/nlu.md
```diff
@@ -4998,28 +4998,34 @@ data/nlu/lookups/products.txt
...
 - hey[^\s]*
 
 ## regex:zipcode
 - [0-9]{5}
+
+## lookup:location.txt
+  /app/git/1/data/location.txt
+
+## lookup:products.txt
+  /app/git/1/data/products.txt
```
erohmensing: This looks incorrect. personally I think the name of the lookup when dumped should just be whatever comes before .txt (though we know it doesn't actually matter for featurization), but the way it dumps the file should be relative so that it works on the server as well as in the repo -- would mention that to Tobi.

PR #459, opened 2020-01-24
Fix build

akelad:
pin pip<20, because of the "skbuild" issue
Related issue: https://github.com/pypa/pip/issues/7629


btotharye:
thanks @akelad merged it as well!

Issue #463, opened 2020-02-04
Bugs in action server


akelad:
```
Attaching to rasa_app_1
app_1              | 2020-01-31 13:03:57 INFO     rasa_sdk.endpoint  - Starting action endpoint server...
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_explain_sales_form'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_chitchat'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_faqs'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_pause'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_store_unknown_product'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_store_unknown_nlu_part'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_store_bot_language'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_store_entity_extractor'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_set_onboarding'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_store_problem_description'.
app_1              | 2020-01-31 13:03:58 INFO     rasa_sdk.executor  - Registered function for 'action_greet_user'.
app_1              | 2020-01-31 13:04:01 INFO     rasa_sdk.executor  - Registered function for 'action_default_ask_affirmation'.
app_1              | 2020-01-31 13:04:01 INFO     rasa_sdk.executor  - Registered function for 'action_default_fallback'.
app_1              | 2020-01-31 13:04:02 INFO     rasa_sdk.executor  - Registered function for 'action_get_community_events'.
app_1              | 2020-01-31 13:04:02 INFO     rasa_sdk.executor  - Registered function for 'action_next_step'.
app_1              | 2020-01-31 13:04:02 INFO     rasa_sdk.executor  - Registered function for 'action_docs_search'.
[...]

Issue #422, opened 2019-12-16
bug in action server


erohmensing:
```
app_1              | Exception occurred while handling uri: 'http://app:5055/webhook'
app_1              | Traceback (most recent call last):
app_1              |   File "/usr/local/lib/python3.6/site-packages/sanic/app.py", line 942, in handle_request
app_1              |     response = await response
app_1              |   File "/app/rasa_sdk/endpoint.py", line 86, in webhook
app_1              |     result = await executor.run(action_call)
app_1              |   File "/app/rasa_sdk/executor.py", line 280, in run
app_1              |     events = action(dispatcher, tracker, domain)
app_1              |   File "/app/rasa_sdk/forms.py", line 532, in run
app_1              |     events.extend(self.submit(dispatcher, temp_tracker, domain))
app_1              |   File "/app/demo/actions.py", line 59, in submit
app_1              |     added_to_list = client.subscribe_user(config.mailchimp_list, email)
app_1              |   File "/app/demo/api.py", line 19, in subscribe_user
app_1              |     list_id, data={"email_address": email, "status": "pending"}
app_1              |   File "/usr/local/lib/python3.6/site-packages/mailchimp3/entities/listmembers.py", line 57, in create
app_1              |     check_email(data['email_address'])
app_1              |   File "/usr/local/lib/python3.6/site-packages/mailchimp3/helpers.py", line 62, in check_email
app_1              |     if not re.match(r".+@.+\..+", email):
app_1              |   File "/usr/local/lib/python3.6/re.py", line 172, in match
app_1              |     return _compile(pattern, flags).match(string)
app_1              | TypeError: expected string or bytes-like object
```

erohmensing:
found this on the instance. Looks like somehow, whatever got passed to mailchimp as an email wasn't a string. Which seems super strange, but we should not validate non-strings, and probably do the regex compilation on our side as well (duckling should take care of that but can't be too safe)

PR #464, opened 2020-02-06
Bug in action server(#463)

ArjaanBuijk:
Fixes: 
- https://github.com/RasaHQ/rasa-demo/issues/463
- https://github.com/RasaHQ/rasa-demo/issues/422

All are basically protection against very rare edge cases.

### DIFF
--- a/demo/actions.py
+++ b/demo/actions.py
@@ -684,7 +684,6 @@ class ActionNextStep(Action):
             # trigger get_started_step1 intent
             return [
                 ActionExecuted("action_listen"),
-                SlotSet("step", "1"),
                 UserUttered(
                     "/get_started_step1",
                     {
@@ -710,23 +709,26 @@ class ActionDocsSearch(Action):
         return "action_docs_search"
 
     def run(self, dispatcher, tracker, domain):
+        docs_found = False
         search_text = tracker.latest_message.get("text")
-        # If we're in a TwoStageFallback we need to look back one more user utterance to get the actual text
-        if search_text == "/technical_question{}":
-            last_user_event = get_last_event_for(tracker, "user", skip=2)
-            if last_user_event:
-                search_text = last_user_event.get("text")
-            else:
-                dispatcher.utter_message(text="Sorry, I can't answer your question.")
-                return []
 
         # Search of docs pages
+        alg_res = None
         algolia = AlgoliaAPI(
             config.algolia_app_id, config.algolia_search_key, config.algolia_docs_index
         )
-        alg_res = algolia.search(search_text)
+        if search_text == "/technical_question{}":
+            # If we're in a TwoStageFallback we need to look back one more user utterance
+            # to get the actual text
+            last_user_event = get_last_event_for(tracker, "user", skip=2)
+            if last_user_event:
+                search_text = last_user_event.get("text")
+                alg_res = algolia.search(search_text)
+        else:
+            alg_res = algolia.search(search_text)
 
         if alg_res and alg_res.get("hits") and len(alg_res.get("hits")) > 0:
+            docs_found = True
             doc_list = algolia.get_algolia_link(alg_res.get("hits"), 0)
             doc_list += (
                 "\n" + algolia.get_algolia_link(alg_res.get("hits"), 1)
@@ -738,15 +740,16 @@ class ActionDocsSearch(Action):
                 text="I can't answer your question directly, but I found the following from the docs:\n"
                 + doc_list
             )
-            return [SlotSet("docs_found", True)]
 
-        dispatcher.utter_message(
-            text=(
-                "I can't answer your question directly, and also "
-                "found nothing in our documentation that would help."
+        else:
+            dispatcher.utter_message(
+                text=(
+                    "I can't answer your question directly, and also "
+                    "found nothing in our documentation that would help."
+                )
             )
-        )
-        return [SlotSet("docs_found", False)]
+
+        return [SlotSet("docs_found", docs_found)]
 
 
 class ActionForumSearch(Action):
--- a/demo/actions.py
+++ b/demo/actions.py
@@ -662,12 +662,13 @@ class ActionNextStep(Action):
         return "action_next_step"
 
     def run(self, dispatcher, tracker, domain) -> List[EventType]:
-        step = int(tracker.get_slot("step")) + 1
+        if tracker.get_slot("step"):
+            step = int(tracker.get_slot("step")) + 1
 
-        if step in [2, 3, 4]:
-            dispatcher.utter_message(template=f"utter_continue_step{step}")
-        else:
-            dispatcher.utter_message(template="utter_no_more_steps")
+            if step in [2, 3, 4]:
+                dispatcher.utter_message(template=f"utter_continue_step{step}")
+            else:
+                dispatcher.utter_message(template="utter_no_more_steps")
 
         return []
 
--- a/demo/actions.py
+++ b/demo/actions.py
@@ -12,6 +12,8 @@ from rasa_sdk.events import (
     UserUtteranceReverted,
     ConversationPaused,
     EventType,
+    ActionExecuted,
+    UserUttered,
 )
 from demo.api import MailChimpAPI
 from demo.algolia import AlgoliaAPI
@@ -676,7 +678,21 @@ class ActionNextStep(Action):
             else:
                 dispatcher.utter_message(template="utter_no_more_steps")
 
-        return []
+            return []
+
+        else:
+            # trigger get_started_step1 intent
+            return [
+                ActionExecuted("action_listen"),
+                SlotSet("step", "1"),
+                UserUttered(
+                    "/get_started_step1",
+                    {
+                        "intent": {"name": "get_started_step1", "confidence": 1.0},
+                        "entities": {},
+                    },
+                ),
+            ]
 
 
 def get_last_event_for(tracker, event_type: Text, skip: int = 0) -> Optional[EventType]:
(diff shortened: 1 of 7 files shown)


ArjaanBuijk:
@erohmensing ,
I didn't wait for testing of stories to finish before requesting your review. I hope they finish OK...

If you like the fix for #422, we should use the same test for the emails in the other forms. I did not add it there yet. First wanted to see what you thought of this implementation.

demo/actions.py
```diff
@@ -662,12 +668,13 @@ def name(self) -> Text:
         return "action_next_step"
 
     def run(self, dispatcher, tracker, domain) -> List[EventType]:
-        step = int(tracker.get_slot("step")) + 1
+        if tracker.get_slot("step"):
```
erohmensing: What should this do if there is no step though? Certainly not nothing -- maybe step 1?

demo/actions.py
```diff
@@ -693,24 +700,28 @@ def run(self, dispatcher, tracker, domain):
...
         )
         alg_res = algolia.search(search_text)
 
-        doc_list = algolia.get_algolia_link(alg_res.get("hits"), 0)
-        doc_list += (
-            "\n" + algolia.get_algolia_link(alg_res.get("hits"), 1)
-            if len(alg_res.get("hits")) > 1
-            else ""
-        )
+        if alg_res:
```
erohmensing: If not alg_res, we shouldn't return [], cause it will say nothing and then ask if that helped. Ideally we'd jump straight to the forum search

demo/actions.py
```diff
@@ -724,15 +735,23 @@ def run(self, dispatcher, tracker, domain):
...
+                return []
 
         # Search forum
         discourse = DiscourseAPI("https://forum.rasa.com/search")
         doc_list = discourse.query(search_text)
         doc_list = doc_list.json()
 
-        forum = discourse.get_discourse_links(doc_list.get("topics"), 0)
-        forum += "\n" + discourse.get_discourse_links(doc_list.get("topics"), 1)
+        if doc_list:
```
erohmensing: if not doc_list (also can we please rename this bc i thought this was the docs search lol), can we then also say sth like "sorry, i don't know how to answer that yet" ?

demo/api.py
```diff
@@ -12,6 +13,16 @@ def __init__(self, api_key: Text) -> None:
...
+    @staticmethod
+    def is_valid_email(email: Text) -> bool:
+        """Use mailchimp3 helper function to validate that it will accept it as a valid
+        email"""
+        try:
+            check_email(email)
+            return True
+        except (TypeError, ValueError):
+            return False
+
```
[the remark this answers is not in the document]
erohmensing: mm i see you ran into isuses with flake8 and bare except, i think it's fine if we do that and make flake8 ignore that error, as long as we are logging the error so we can check it out later.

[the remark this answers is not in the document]
ArjaanBuijk: I made the exception handling more precise and issue logging error.

demo/actions.py
ArjaanBuijk: Good idea!

When this happens, I am now triggering the `get_started_step1` intent, and set the slot `step=1`. Works nice.

You can test this fix by starting `rasa shell`, and passing `/next_step`

demo/actions.py
ArjaanBuijk: I renamed `doc_list` and also now give a message with recommendation that they post their question on the forum.

demo/actions.py
ArjaanBuijk: I fixed this as well, by introducing a new slot of type boolean `docs_found`, and used these to distinquish the stories for technical_question intent. If docs are found, it behaves as before. If no docs are found, it will now automatically go and do the forum search.

ArjaanBuijk:
@erohmensing , all your feedback was addressed. Let me know if it looks good.

---
The logger will print this type of warning if check_email raises an Exception. I passed an integer while in the debugger to produce this:
```
2020-02-06 14:03:59 WARNING  demo.api  - Error: exception in check_email.
<class 'TypeError'> - expected string or bytes-like object
email = 1
type(email) = <class 'int'>
```

---
The updates for the technical_question is based on a new slot `docs_found` with updated stories.

---
The protection against someone sending /action_next_step while no step slot is filled yet is based on something I found in the [forum](https://forum.rasa.com/t/trigger-a-story-or-intent-from-a-custom-action/13784/9). Works great! And it is a really good example why it is better to trigger intents than actions. The learned stories kick in seamlessly.

demo/actions.py
```diff
@@ -662,14 +670,29 @@ def name(self) -> Text:
...
+        else:
+            # trigger get_started_step1 intent
+            return [
+                ActionExecuted("action_listen"),
+                SlotSet("step", "1"),
+                UserUttered(
+                    "/get_started_step1",
+                    {
+                        "intent": {"name": "get_started_step1", "confidence": 1.0},
+                        "entities": {},
```
erohmensing: How does the step slot get set usually? Would it normally happen after the /get_started_step1 intent?

demo/discourse.py
```diff
@@ -9,9 +9,11 @@ def __init__(self, url: Text):
...
 
     @staticmethod
-    def get_discourse_links(topics: List[Dict[Text, Any]], index: int):
-        doc_url = f"https://forum.rasa.com/t/{topics[index]['slug']}/{str(topics[index]['id'])}"
-        forum = f"- [{topics[index]['title']}]({doc_url})"
+    def get_discourse_links(topics: Optional[List[Dict[Text, Any]]], index: int):
+        forum = None
+        if topics:
+            doc_url = f"https://forum.rasa.com/t/{topics[index].get('slug')}/{str(topics[index].get('id'))}"
+            forum = f"- [{topics[index]['title']}]({doc_url})"
```
erohmensing:
```suggestion
            forum = f"- [{topics[index].get('title')}]({doc_url})"
```
would you mind doing a search for any more of these?

demo/actions.py
```diff
@@ -693,26 +716,37 @@ def run(self, dispatcher, tracker, domain):
...
+                else ""
+            )
+
+            dispatcher.utter_message(
+                text="I can't answer your question directly, but I found the following from the docs:\n"
+                + doc_list
+            )
+            return [SlotSet("docs_found", True)]
 
         dispatcher.utter_message(
```
erohmensing: I know it's the same logically, but can we put this in an `else:` to make it clearer off the bat?

erohmensing: Do we want to actually say this though? i think we want it to be quiet about the fact that it didn't find stuff in the docs and go directly to the forum search as a first response instead. right?

demo/api.py
```diff
@@ -12,6 +16,24 @@ def __init__(self, api_key: Text) -> None:
...
+        except ValueError:  # purposely raised in case of invalid email
+            return False
+        except Exception as e:
+            logger.warning(
+                f"Error: exception in check_email.\n"
+                f"{type(e)} - {e}\n"
+                f"email = {email}\n"
+                f"type(email) = {type(email)}"
+            )
+            return False
```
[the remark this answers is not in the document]
ArjaanBuijk: Yeah, good idea to str() it within the try: block before sending it to check_email.

The rest of the code already does exactly the same as what you propose, except (no pun intended! ) that I am not using a `bare except`, because that it is a bad practice with unforeseen consequences, like not being able to CTRL+C applications if you are not careful.

The message I am giving will allow us to see what was passed. Just doing logger.error(e) is not giving the actual value or type of what was passed, so it will be hard to find out where a weird value came from.

What do you think?

demo/actions.py
ArjaanBuijk: Normally it happens like this:

```
* get_started_step1
    - action_greet_user
```

The custom action `action_greet_user` then sets the step:

```
    elif intent[:-1] == "get_started_step" and shown_privacy:
            dispatcher.utter_message(template=f"utter_{intent}")
            return [SlotSet("step", intent[-1])]
```

sooooo, I think I do not need to set the step myself. It should happen automatic. Let me test this out.

demo/discourse.py
ArjaanBuijk: Yep!  I will do a search on these type of cases.

demo/api.py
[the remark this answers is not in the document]
erohmensing: Yes, definitely except Exception as e, mine was just pseudocode 😅

The reason I was thinking to keep it more general instead of printing the types, is that if we end up hitting something other than a type error, the printed information might not give us very helpful information. We could also do both.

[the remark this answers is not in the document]
erohmensing: Ahh i see, i think i was reading it wrong. here it still gives the type of the error and the output? in that case, yes, feel free to leave as is. 👍 sorry about that

demo/actions.py
erohmensing: Hm maybe the action sets it, it just isn't featurized and therefore isn't in the stories? haven't looked too deeply into the stories/actions.

demo/actions.py
ArjaanBuijk: About the `else`, I will refactor it a little bit to get rid of the multiple return statements. That always makes it hard to read and dangerous to have logic fall through to the bottom without a proper return.

About the message when no docs are found. When I was testing it, I found it much better to provide consistent feedback, to avoid confusion.

If for some questions Sara starts talking about the docs, and for others only about the forum, as a user I am wondering why it is not telling me about the docs for some questions. It appears like a bug.

erohmensing: Hm, interesting, I woudn't see it that way, but I guess that makes sense. You've been the one testing it, so I trust your UX gut feeling.

demo/actions.py
ArjaanBuijk: @erohmensing ,
I can indeed remove it, and then the slot `step` will be set as usual by the `action_greet_user`.

Good catch!

Issue #468, opened 2020-02-25
use rasa x tags to track thumbs up/down


erohmensing:
I think we currently have 2 areas with thumbs up/down for feedback purposes:

- General utter_ask_feedback or something like that
- Was this helpful? After docs search 

We should use the Rasa X tags API to auto-tag those conversations so we can easily filter for good and bad feedback. 

Another q (different issue if we did): should we add another "was that helpful?" after forum search? Discourse's search API isn't great, so i'm wondering if knowing whether the links are helpful or not is something we should look at.

PR #477, opened 2020-03-13
conversation tags for feedback

melindaloubser1:
Closes https://github.com/RasaHQ/rasa-demo/issues/468
Add automatic conversation tags for positive/negative feedback

### DIFF
--- a/demo/actions.py
+++ b/demo/actions.py
@@ -26,10 +26,6 @@ from demo.gdrive_service import GDriveService
 
 logger = logging.getLogger(__name__)
 
-# RASA_X_HOST = "rasa-x:5002"
-
-RASA_X_HOST = "localhost:5002"
-
 
 class SubscribeNewsletterForm(FormAction):
     """Asks for the user's email, call the newsletter API and sign up user"""
@@ -799,22 +795,13 @@ class ActionForumSearch(Action):
         return []
 
 
-class ActionTagConvo(Action):
-    """Returns the chitchat utterance dependent on the intent"""
-
-    def name(self):
-        return NotImplementedError
-
-    def tag_convo(self, tracker, label):
-        endpoint = f"http://{RASA_X_HOST}/api/conversations/{tracker.sender_id}/tags"
-        requests.post(url=endpoint, data=label)
-
-    def run(self, dispatcher, tracker, domain) -> List[EventType]:
-        return []
+def tag_convo(tracker, label):
+    endpoint = f"http://{config.rasa_x_host}/api/conversations/{tracker.sender_id}/tags"
+    requests.post(url=endpoint, data=label)
 
 
-class ActionTagFeedback(ActionTagConvo):
-    """Returns the chitchat utterance dependent on the intent"""
+class ActionTagFeedback(Action):
+    """Tag a conversation in Rasa X as positive or negative feedback """
 
     def name(self):
         return "action_tag_feedback"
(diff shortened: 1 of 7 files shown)


melindaloubser1:
I've implemented it as two separate actions, but could also have it be one action based on the "feedback" slot (which we could set after the docs search question too).

erohmensing:
Oh lol, did not see your comment. Yeah i think using feedback slot for the regular feedback one makes sense if we both agree. but i'd prefer to keep it separate, bc the docs could return a bad search but that doesn't mean that the interaction was necessarily bad

melindaloubser1:
sure, that makes sense, I'll split it into four tags (three actions then)

melindaloubser1:
@erohmensing I fixed the story errors; could you take another look?

demo/actions.py
```diff
@@ -25,6 +26,10 @@
 
 logger = logging.getLogger(__name__)
 
+# RASA_X_HOST = "rasa-x:5002"
+
+RASA_X_HOST = "localhost:5002"
+
```
erohmensing: Let's make this a config parameter in demo/config.py. then we can set it in the docker-compose override

demo/actions.py
```diff
@@ -792,3 +797,60 @@ def run(self, dispatcher, tracker, domain):
...
+    def name(self):
+        return NotImplementedError
+
+    def tag_convo(self, tracker, label):
+        endpoint = f"http://{RASA_X_HOST}/api/conversations/{tracker.sender_id}/tags"
+        requests.post(url=endpoint, data=label)
+
+    def run(self, dispatcher, tracker, domain) -> List[EventType]:
+        return []
+
```
erohmensing: I think we can just make this a `tag_convo` helper method outside of all of the action classes, can't we? instead of making a new action class for it

domain.yml
```diff
@@ -1,6 +1,6 @@
 session_config:
-  session_expiration_time: 480  # value in minutes, 0 means infinitely long
-  carry_over_slots_to_new_session: true  # set to false to forget slots between sessions
+  session_expiration_time: 480.0
+  carry_over_slots_to_new_session: true
 intents:
 - affirm
 - ask_builder
```
erohmensing: Looks like we ddin't change anything to the domain, rasa x did -- let's just remove the diff bc it looks to be screwing up e.g. the retrieval intents

demo/actions.py
```diff
@@ -792,3 +797,60 @@ def run(self, dispatcher, tracker, domain):
...
+        else:
+            return []
+
+        self.tag_convo(tracker, label)
+
+        return []
+
+
+class ActionTagDocsSearch(ActionTagConvo):
+    """Returns the chitchat utterance dependent on the intent"""
```
erohmensing: oops, please also give each of these methods a correct doc string

domain.yml
melindaloubser1: urg i undid that so many times and still ...

demo/actions.py
melindaloubser1: sure, that works too

domain.yml
erohmensing: Hahaha. I guess just at the very end 😄do you always `git add .`? you can also `git add folder/` or `git add path/to/filename.ext`

melindaloubser1: no I usually `git add -u` or specific folders/filenames. Since domain had to get new action names I didn't realize other changes also got committed

demo/actions.py
```diff
@@ -792,3 +793,51 @@ def run(self, dispatcher, tracker, domain):
             )
 
         return []
+
+
+def tag_convo(tracker, label):
```
erohmensing: types and a docstring here please

demo/actions.py
```diff
@@ -792,3 +793,51 @@ def run(self, dispatcher, tracker, domain):
...
+        return []
+
+
+class ActionTagDocsSearch(Action):
+    """Tag a conversation in Rasa X according to whether the docs search was helpful"""
+
+    def name(self):
+        return "action_tag_docs_search"
+
+    def run(self, dispatcher, tracker, domain) -> List[EventType]:
```
erohmensing: are the other action's `run()` methods typed? if so, let's add types, if not, don't worry about it

melindaloubser1: the method returns are typed, but not the arguments

Issue #481, opened 2020-03-30
learn more about 'rasa open source' leads to `ask_faq_opensource_cost`


erohmensing:
see http://website-demo.rasa.com/conversations/ef8c2fa8da2f470a8a67d4a3f7ab0444/1585325487.726087

and the slack thread here: https://rasa-hq.slack.com/archives/C4Q4X95M4/p1585309557028800

akelad:
@melindaloubser1 do you have time to work on this? I think it should be a quick fix

melindaloubser1:
yes I should be able to get to it this week

PR #482, opened 2020-04-02
Fix onboarding flow bug with how to get started with rasa oss

melindaloubser1:
- closes https://github.com/RasaHQ/rasa-demo/issues/481
- Add training data for "rasa open source" to the `how_to_get_started` intent
- add more stories with `how_to_get_started{"product": "rasa"}`

### DIFF
--- a/data/nlu/nlu.md
+++ b/data/nlu/nlu.md
@@ -3765,6 +3765,14 @@
 - where to start?
 - yes i wanna know more about rasa [nlu](product)
 - I am trying to build a bot using rasa
+- more about [Rasa open source](product)
+- tell me about [rasa open source](product)
+- i want to know about [rasa Open source](product)
+- get started with [Rasa Open Source](product)
+- [Rasa open source](product)
+- [rasa open source](product)
+- [rasa Open source](product)
+- [Rasa Open Source](product)
 
 ## intent:human_handoff
 - Can I speak to anyone who can really help me?
@@ -5014,6 +5022,10 @@ data/nlu/lookups/products.txt
 
 ## synonym:rasa
 - Rasa
+- rasa open source
+- Rasa Open Source
+- Rasa open source
+- rasa Open Source
 
 ## synonym:stack
 - fullstack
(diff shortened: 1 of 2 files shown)


data/nlu/nlu.md
```diff
@@ -3765,6 +3765,14 @@
 - where to start?
 - yes i wanna know more about rasa [nlu](product)
 - I am trying to build a bot using rasa
```
akelad, line 3767: we should probably annotate rasa as a product as well then everywhere.

data/nlu/nlu.md
```diff
@@ -3765,6 +3765,14 @@
...
+- more about [Rasa open source](product)
+- tell me about [rasa open source](product)
+- i want to know about [rasa Open source](product)
+- get started with [Rasa Open Source](product)
+- [Rasa open source](product)
+- [rasa open source](product)
+- [rasa Open source](product)
+- [Rasa Open Source](product)
```
akelad, line 3775: did you check this doesn't cause conflict with `ask_faq_rasaoss` or whatever the faq intent about RAsa open source is

melindaloubser1: In evaluation it does not, it can still differentiate between "is rasa open source" and "rasa open source".

data/nlu/nlu.md
melindaloubser1: It seems like `rasa` is used a lot though in contexts where it isn't about the oss product - maybe better to turn `rasa open source` into its own entity? Since we started referring to it that way more recently

data/nlu/nlu.md
akelad: interesting 🤔

data/nlu/nlu.md
akelad: hm ok. but then i would say leaving the entity annotated as it is now is fine, will approve the PR

melindaloubser1: maybe if we notice a change in how users use it in conversations later we can change it

PR #486, opened 2020-04-21
Fix FallbackAction behaviour with out_of_scope

akelad:


akelad:
fyi @melindaloubser1 this is the fix i was talking about

akelad:
@erohmensing any ideas why flake8 is complaining here? https://github.com/RasaHQ/rasa-demo/runs/609045803#step:5:1
To me it's an incorrect complaint, and also contradicts black

erohmensing:
Wait, black has you add those commas at the end even when it's one line? that doesn't seem (or look) right

data/nlu/nlu.md
```diff
@@ -4409,6 +4420,14 @@
 - you will know it from the single red rose it carries between its teeth
 - you're a woman
 - ı am learning python
+- what kind of bird are you?
+- contextua
+- contextual
+- talk to me about voulette
+- voulette voulette
+- Do you have a demo?
```
erohmensing, line 4428: We should keep an eye on if this is a common thing 🤔

data/nlu/nlu.md
```diff
@@ -5054,3 +5070,9 @@ data/nlu/lookups/products.txt
 
 ## regex:zipcode
 - [0-9]{5}
+
+## lookup:location.txt
+  data/nlu/lookups/location.txt
+
+## lookup:products.txt
+  data/nlu/lookups/products.txt
```
erohmensing, line 5078: This is annoying and Rasa X shouldn't be doing this IMO 👀not sure if you want to merge it or not (the features are the same regardless of the lookup name right?

actions/actions.py
```diff
@@ -484,8 +484,11 @@ def run(
                 intent_ranking = intent_ranking[:1]
         first_intent_names = [
             intent.get("name", "")
+            if intent.get("name", "") not in ["out_of_scope", "faq", "chitchat"]
+            else tracker.latest_message.get("response_selector")
+            .get(intent.get("name", ""))
+            .get("full_retrieval_intent")
```
erohmensing, line 490: maybe it makes sense to make a little helper method for this that's more intuitive with a proper if/else so that the next person trying to fix this isn't gonna want to grumpily @ you 

Also is there an easy way for us to automate getting a list of response selector intents so we don't have to update that? maybe we can reuse some rasa code?

erohmensing: hm but then again we don't want to have to import the huge rasa package to do so

actions/actions.py
```diff
@@ -501,22 +504,18 @@ def run(
         for intent in first_intent_names:
             logger.debug(intent)
             logger.debug(entities)
-            buttons.append(
-                {
-                    "title": self.get_button_title(intent, entities),
-                    "payload": "/{}{}".format(intent, entities_json),
-                }
-            )
+            if len(intent.split("/")) > 1:
```
erohmensing, line 507: can we just do `if "/" in intent`?

actions/actions.py
```diff
@@ -501,22 +504,18 @@ def run(
...
+                message = self.get_button_title(intent, entities)
+                buttons.append({"title": message, "payload": message})
+            else:
+                buttons.append(
+                    {
+                        "title": self.get_button_title(intent, entities),
```
erohmensing, line 513: this "title" is actually the same as L 508-509 but it doesn't look so, let's pull out 508 outside of if/else and use it in both titles

actions/actions.py
```diff
@@ -501,22 +504,18 @@ def run(
...
             logger.debug(entities)
-            buttons.append(
-                {
-                    "title": self.get_button_title(intent, entities),
-                    "payload": "/{}{}".format(intent, entities_json),
-                }
-            )
+            if len(intent.split("/")) > 1:
+                message = self.get_button_title(intent, entities)
+                buttons.append({"title": message, "payload": message})
```
erohmensing, line 509: add a comment about why we have to use the message instead of regex format for retrieval intents

actions/actions.py
```diff
@@ -501,22 +504,18 @@ def run(
...
+            else:
+                buttons.append(
+                    {
+                        "title": self.get_button_title(intent, entities),
+                        "payload": "/{}{}".format(intent, entities_json),
```
erohmensing, line 514:
```suggestion
                        "payload": f"/{intent}{entities_json}",
```

actions/actions.py
```diff
@@ -501,22 +504,18 @@ def run(
         for intent in first_intent_names:
             logger.debug(intent)
             logger.debug(entities)
```
erohmensing, line 506: Let's remove these or make a more useful debug message if we feel we need it

erohmensing:
Also please give it a helpful PR name 🙂

data/nlu/nlu.md
akelad: i mean, Rasa X will continue doing it so I would just merge it, the name doesn't matter

actions/actions.py
akelad: lol smart

akelad:
```suggestion
            if "/" in intent:
```

actions/actions.py
akelad: yeah let's not :D not sure it's worth making a helper method, but I can leave a comment on what is happening here

akelad:
@erohmensing you can try it out if you want, but I did try it out 👀 just not with a new rasa model yet, i'm gonna wait with that when everything is in master, including melindas PR

akelad:
@erohmensing approve pls 🙏

actions/actions.py
erohmensing: I think a helper method would be better as it wouldn't necessitate a 4 line comment, but whatever 😄i might end up changing it next time i edit the actions though 😉

akelad: then the helper method would have a 4 line comment 😂

Issue #479, opened 2020-03-23
Make training workflows in actions conditional 


akelad:
We should only run NLU evaluation/Core model testing when those files change as it takes up a lot of time and can become costly.
Options:
* Check whether the files have changed
* Only run certain workflows if a label is present on a PR

PR #488, opened 2020-04-22
Conditional training/testing steps if files changed

melindaloubser1:
Closes https://github.com/RasaHQ/rasa-demo/issues/479
Makes training & testing steps conditional on whether certain files were changed.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -128,29 +128,29 @@ jobs:
       if: |
         contains(  steps.files.outputs.all, ' actions/' ) 
         || contains(  steps.files.outputs.all, ' Dockerfile' )
-      run: echo "::set-env name=ACTIONS::true"
+      run: echo "::set-env name=ACTIONS_CHANGED::true"
     - name: Authenticate into Google Cloud Platform
-      if: env.ACTIONS == 'true'
+      if: env.ACTIONS_CHANGED == 'true'
       uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
       with:
         version: '275.0.0'
         service_account_key: ${{ secrets.GCLOUD_AUTH }}
     - name: Configure Docker to use Google Cloud Platform
-      if: env.ACTIONS == 'true'
+      if: env.ACTIONS_CHANGED == 'true'
       run: |
         gcloud auth configure-docker
     - name: Pull Latest Image
-      if: env.ACTIONS == 'true'
+      if: env.ACTIONS_CHANGED == 'true'
       run: |
         docker pull gcr.io/replicated-test/rasa-demo:latest || true
     - name: Set Build ID from PR Ref
-      if: env.ACTIONS == 'true'
+      if: env.ACTIONS_CHANGED == 'true'
       run: echo ::set-env name=BUILD_NUMBER::$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
     - name: Build Image
-      if: env.ACTIONS == 'true'
+      if: env.ACTIONS_CHANGED == 'true'
       run: |
         docker build -t gcr.io/replicated-test/rasa-demo:pr$BUILD_NUMBER -t gcr.io/replicated-test/rasa-demo:latest --cache-from gcr.io/replicated-test/rasa-demo:latest .
     - name: Push PR Image to Google Cloud Container Registry
-      if: env.ACTIONS == 'true'
+      if: env.ACTIONS_CHANGED == 'true'
       run: |
         docker push gcr.io/replicated-test/rasa-demo:pr$BUILD_NUMBER
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -122,24 +122,29 @@ jobs:
     - id: files
       uses: jitterbit/get-changed-files@v1
     - name: Authenticate into Google Cloud Platform
+      if: |
+        contains(  steps.files.outputs.all, 'actions/' ) 
+        || contains(  steps.files.outputs.all, 'Dockerfile' )
       uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
       with:
         version: '275.0.0'
         service_account_key: ${{ secrets.GCLOUD_AUTH }}
-      if: |
-        contains(  steps.files.outputs.all, 'actions/' ) 
-        || contains(  steps.files.outputs.all, 'Dockerfile' )
     - name: Configure Docker to use Google Cloud Platform
+      if: success()
       run: |
         gcloud auth configure-docker
     - name: Pull Latest Image
+      if: success()
       run: |
         docker pull gcr.io/replicated-test/rasa-demo:latest || true
     - name: Set Build ID from PR Ref
+      if: success()
       run: echo ::set-env name=BUILD_NUMBER::$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
     - name: Build Image
+      if: success()
       run: |
         docker build -t gcr.io/replicated-test/rasa-demo:pr$BUILD_NUMBER -t gcr.io/replicated-test/rasa-demo:latest --cache-from gcr.io/replicated-test/rasa-demo:latest .
     - name: Push PR Image to Google Cloud Container Registry
+      if: success()
       run: |
         docker push gcr.io/replicated-test/rasa-demo:pr$BUILD_NUMBER
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -3,6 +3,34 @@ name: Build and Deploy
 on: [pull_request]
 
 jobs:
+  set-changed-files:
+    name: Set Changed Files Env Vars
+    runs-on: ubuntu-latest
+    steps:
+    - id: files
+      uses: jitterbit/get-changed-files@v1
+    - name: set_training
+      if: |
+          contains(  steps.files.outputs.all, 'data/' ) 
+          || contains(  steps.files.outputs.all, 'config.yml' ) 
+          || contains(  steps.files.outputs.all, 'domain.yml' )
+      run: echo "::set-env name=TRAINING::true"
+    - name: set_nlu
+      if: |
+        contains(  steps.files.outputs.all, 'data/nlu/' ) 
+        || contains(  steps.files.outputs.all, 'config.yml' )
+      run: echo "::set-env name=NLU::true"
+    - name: set_stories
+      if: |
+        ( contains(  steps.files.outputs.all, 'data/core/' ) 
+        || contains(  steps.files.outputs.all, 'config.yml' ) 
+        || contains(  steps.files.outputs.all, 'domain.yml' ) )
+      run: echo "::set-env name=STORIES::true"
+    - name: set_actions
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -59,36 +59,57 @@ jobs:
...
       run: |
         rasa test core --stories test/test_stories.md --fail-on-prediction-errors
     - name: Upload model
-      if: github.ref == 'refs/heads/master'
+      if: |
+          github.ref == 'refs/heads/master'
+          && steps.train.outcome == 'success'
       uses: actions/upload-artifact@master
       with:
         name: model
```
erohmensing, line 115: What do you think about action server images, should they only be built if the dockerfile or sth in the `actions/` folder changes?

.github/workflows/build_and_deploy.yml
```diff
@@ -59,36 +59,57 @@ jobs:
...
-        pip install -r requirements-dev.txt
+          python -m pip install --upgrade "pip<20"
+          pip install -r requirements-dev.txt
     - name: Cross-validate NLU model
-      run: |
-        rasa test nlu -f 3 --cross-validation
-        python .github/workflows/format_results.py
+      if: |
+          contains(  steps.files.outputs.all, 'data/nlu/' ) 
+          || contains(  steps.files.outputs.all, 'config.yml' )
```
erohmensing, line 80: still wondering if we should try to do NLU on a specific tag, bc anytime we add rasa x nlu annotations it will trigger this (actually anything from rasa x rn bc of the synonyms re-ordering bug)
though i think changes to the config make sense as always triggering it

.github/workflows/build_and_deploy.yml
```diff
@@ -59,36 +59,57 @@ jobs:
...
         rasa train
     - name: Test Training Stories
+      if: |
+          ( contains(  steps.files.outputs.all, 'data/core/' ) 
+          || contains(  steps.files.outputs.all, 'config.yml' ) 
+          || contains(  steps.files.outputs.all, 'domain.yml' ) )
+          && steps.train.outcome == 'success'
       working-directory: ${{ github.workspace }}
       run: |
         rasa test core --stories test/test_stories.md --fail-on-prediction-errors
```
erohmensing, line 108: If our test stories are e2e, we should run them on NLU changes too

.github/workflows/build_and_deploy.yml
melindaloubser1: yeah, I think that makes sense. I'll add that condition

.github/workflows/build_and_deploy.yml
melindaloubser1: Ok, will add that

.github/workflows/build_and_deploy.yml
melindaloubser1: as long as anyone making changes will add the flag when required - since we're not in the habit of it that's the only reason I wonder if it's good to do that?

.github/workflows/build_and_deploy.yml
melindaloubser1: should we keep the core story tests and add the e2e tests? we didn't have any before

.github/workflows/build_and_deploy.yml
erohmensing: I think (hope) that any changes big enough to NLU that would warrant the need for the flag would push the person to want to test those changes, and the point of this was to make it not happen all the time, so i think it's okay. I definitely think config should always cause it though (or at least the pipeline, but dont think we can get that specific about it)

.github/workflows/build_and_deploy.yml
erohmensing: ah do we still not have any e2e tests? lol yikes

erohmensing: in that case we can leave it like this for now

erohmensing:
@melindaloubser1 is this ready for another review?

.github/workflows/build_and_deploy.yml
melindaloubser1: I'll update when the response selector PR is merged, it contains e2e tests

melindaloubser1: ** I mean I'll update it in that PR

.github/workflows/build_and_deploy.yml
melindaloubser1: sure, it's been added =)

melindaloubser1:
@erohmensing yep, added label check `nlu_testing_required`. logic is label &&  nludata change -> run test, should it be label || nludata change?

erohmensing:
looks like it's already OR'd with config change so, i guess `label && nludata change` still makes sense, if both config and nlu data are still the same i don't think i can think of a reason to run it 🤔

melindaloubser1:
Ja me neither.

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - id: files
+      uses: jitterbit/get-changed-files@v1
+    - name: set_actions
+      if: |
+        contains(  steps.files.outputs.all, ' actions/' ) 
+        || contains(  steps.files.outputs.all, ' Dockerfile' )
+      run: echo "::set-env name=ACTIONS::true"
```
erohmensing, line 131:
```suggestion
      run: echo "::set-env name=ACTIONS_CHANGED::true"
```
bc this is open source, might help to make them more descriptive (same with env.TRAINING)

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
...
     - name: Authenticate into Google Cloud Platform
+      if: env.ACTIONS == 'true'
```
melindaloubser1, line 133:
```suggestion
      if: env.ACTIONS_CHANGED == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
...
       uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
       with:
         version: '275.0.0'
         service_account_key: ${{ secrets.GCLOUD_AUTH }}
     - name: Configure Docker to use Google Cloud Platform
+      if: env.ACTIONS == 'true'
```
melindaloubser1, line 139:
```suggestion
      if: env.ACTIONS_CHANGED == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
...
       run: |
         gcloud auth configure-docker
     - name: Pull Latest Image
+      if: env.ACTIONS == 'true'
```
melindaloubser1, line 143:
```suggestion
      if: env.ACTIONS_CHANGED == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
...
       run: |
         docker pull gcr.io/replicated-test/rasa-demo:latest || true
     - name: Set Build ID from PR Ref
+      if: env.ACTIONS == 'true'
```
melindaloubser1, line 147:
```suggestion
      if: env.ACTIONS_CHANGED == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
...
       run: echo ::set-env name=BUILD_NUMBER::$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
     - name: Build Image
+      if: env.ACTIONS == 'true'
```
melindaloubser1, line 150:
```suggestion
      if: env.ACTIONS_CHANGED == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -99,22 +122,35 @@ jobs:
...
       run: |
         docker build -t gcr.io/replicated-test/rasa-demo:pr$BUILD_NUMBER -t gcr.io/replicated-test/rasa-demo:latest --cache-from gcr.io/replicated-test/rasa-demo:latest .
     - name: Push PR Image to Google Cloud Container Registry
+      if: env.ACTIONS == 'true'
```
melindaloubser1, line 154:
```suggestion
      if: env.ACTIONS_CHANGED == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -60,35 +60,58 @@ jobs:
...
     steps:
     - uses: actions/checkout@v1
+    - id: files
+      uses: jitterbit/get-changed-files@v1
+    - name: set_training
+      if: |
+          contains(  steps.files.outputs.all, ' data/' ) 
+          || contains(  steps.files.outputs.all, ' config.yml' ) 
+          || contains(  steps.files.outputs.all, ' domain.yml' )
+      run: echo "::set-env name=TRAINING::true"
```
melindaloubser1, line 70:
```suggestion
      run: echo "::set-env name=RUN_TRAINING::true"
```

.github/workflows/build_and_deploy.yml
```diff
@@ -60,35 +60,58 @@ jobs:
...
     - name: Set up Python 3.7
+      if: env.TRAINING == 'true'
```
melindaloubser1, line 72:
```suggestion
      if: env.RUN_TRAINING == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -60,35 +60,58 @@ jobs:
...
       uses: actions/setup-python@v1
       with:
         python-version: 3.7
     - name: Install dependencies
+      if: env.TRAINING == 'true'
```
melindaloubser1, line 77:
```suggestion
      if: env.RUN_TRAINING == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -60,35 +60,58 @@ jobs:
...
     - name: post cross-val results to PR
+      if: steps.cvnlu.outcome == 'success'
       uses: amn41/comment-on-pr@comment-file-contents
       continue-on-error: true
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         msg: results.md
     - name: Train Model
+      if: env.TRAINING == 'true'
```
melindaloubser1, line 99:
```suggestion
      if: env.RUN_TRAINING == 'true'
```

.github/workflows/build_and_deploy.yml
```diff
@@ -60,35 +60,58 @@ jobs:
...
+          || contains(  steps.files.outputs.all, ' config.yml' ) 
+          || contains(  steps.files.outputs.all, ' domain.yml' )
       working-directory: ${{ github.workspace }}
       run: |
         rasa test core --stories test/test_stories.md --fail-on-prediction-errors
     - name: Upload model
-      if: github.ref == 'refs/heads/master'
+      if: |
+          github.ref == 'refs/heads/master' 
+          && env.TRAINING == 'true'
```
melindaloubser1, line 114:
```suggestion
          && env.RUN_TRAINING == 'true'
```

Issue #470, opened 2020-03-02
migrate from unspecified tools


melindaloubser1:
user input "from which tools can i migrate to Rasa" doesn't trigger `migrate_from` action; it should trigger it and the action should handle a question about migrating from unspecified tech.

melindaloubser1:
@erohmensing locally this works, but deployed sara no go. This (and related questions) triggers fallback, but choosing `switch` just goes back to fallback instead of triggering `utter_no_guide_for_switch`. Do you know why that might be happening in the deployed one?

erohmensing:
Hm when i type that in live on sara, `switch` isn't even an option in the fallback

akelad:
@melindaloubser1 is this still blocked?

melindaloubser1:
@akelad  I saw it in live sara again today, will try to replicate it locally again - that was the blocker, could not replicate locally.

melindaloubser1:
observed this behavior in rasa x 
```
## New Story
* switch
    - action_store_unknown_product
    - slot{"unknown_product":"/switch"}
    - utter_no_guide_for_switch
    - utter_anything_else
* switch{"current_api":"tensorflow"}
    - slot{"current_api":"tensorflow"}
    - utter_switch_luis
    - utter_anything_else
* switch{"current_api":"tensorflow"}
    - slot{"current_api":"tensorflow"}
    - utter_switch_dialogflow
    - utter_anything_else
* switch{"current_api":"tensorflow"}
    - slot{"current_api":"tensorflow"}
    - utter_switch_dialogflow
    - utter_anything_else
```

akelad:
ok something is wrong with that indeed haha

melindaloubser1:
https://website-demo.rasa.com/conversations/000f249134ac4818b2c20e4c507e1b44

melindaloubser1:
@akelad I still can't reproduce this locally, nor can I reproduce it in rasa x atm. looking at the trackers from the weird convos to see what went wrong there.

akelad:
ok so adding a story for other should fix it right? Because then we just tell them there is no migration guide right?
Or what do you mean with it not explaining the conversation above?

erohmensing:
I think she means that if you just have /switch with no current_api we should then respond with "what tool are you migrating from?"

melindaloubser1:
[...]

PR #492, opened 2020-04-29
Migrate from other tool

melindaloubser1:
- Closes https://github.com/RasaHQ/rasa-demo/issues/470
- In the migration flow, handle unknown tools


data/core/step1_get_started.md
```diff
@@ -180,6 +180,32 @@
     - utter_no_guide_for_switch
     - utter_anything_else
 
+
+## just switch
+* switch{"current_api":"tensorflow"}
+    - slot{"current_api":"__other__"}
```
ArjaanBuijk, line 186: Does `__other__` have a special meaning, compared to `other` ?

melindaloubser1: yes, it's the reserved key for categorical slot values that aren't specified in the domain

PR #493, opened 2020-04-29
Rasa X annotations

akelad:


data/nlu/nlu.md
```diff
@@ -3934,9 +3946,8 @@
 - please tell steps for installing chatbot
 - what do I need to install Rasa
 - what sould i do to install rasa
+- installation steps of rasa
 - how can I install [rasa open source](product)?
-- The [open source](product) version
-- what is [rasa open source](product) used for?
```
akelad, line 3939: @melindaloubser1 it looks like you added these, was that an accident?

melindaloubser1: I think  I added them because for the getting started flow these were input as answers to "which one do you want to know more about?" and for that to work the product has to be extracted

akelad: yes but i don't think it should be part of the installation intent :D

melindaloubser1: eek no they should not! my bad

data/nlu/nlu.md
```diff
@@ -4012,6 +4024,7 @@
 - please tell steps for installing chatbot
 - what do I need to install Rasa
 - what sould i do to install rasa
+- installation steps of rasa
 - how can I install [rasa open source](product)?
 - The [open source](product) version
 - what is [rasa open source](product) used for?
```
melindaloubser1, line 4030:
```suggestion
```

Issue #496, opened 2020-04-30
Fix hacky "action_next_step"


akelad:
Forcing a user message/flow by adding events to the tracker isn't really recommended. We should find a better way to do this: https://github.com/RasaHQ/rasa-demo/blob/8b1ba1b2843b6844ad57fa7361f67c0a7aff9576/actions/actions.py#L645-L656

PR #497, opened 2020-04-30
Fix hacky next_step handling

melindaloubser1:
- closes https://github.com/RasaHQ/rasa-demo/issues/496
- Use a follow up action instead of a user utterance to trigger `action_greet_user` for `/next_step{"step":null}`

### DIFF
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -13,8 +13,7 @@ from rasa_sdk.events import (
     UserUtteranceReverted,
     ConversationPaused,
     EventType,
-    ActionExecuted,
-    UserUttered,
+    FollowupAction,
 )
 
 from actions import config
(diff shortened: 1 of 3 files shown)


actions/actions.py
```diff
@@ -643,17 +644,7 @@ def run(self, dispatcher, tracker, domain) -> List[EventType]:
...
-            return [
-                ActionExecuted("action_listen"),
-                UserUttered(
-                    "/get_started_step1",
-                    {
-                        "intent": {"name": "get_started_step1", "confidence": 1.0},
-                        "entities": [],
-                    },
-                ),
-            ]
```
erohmensing, line 656: I'm confused about this, does `action_greet_user` usually happen after `get_started_step1`? shouldn't it go to the beginning of the onboarding flow?

melindaloubser1: Yeah, I was confused too, but it looks like this is a kinda rare scenario where the user says "next step", before any steps have been completed, or maybe after all steps have been completed.

melindaloubser1: e.g. start the conversation with "hi", then immediately after write "next step please"

melindaloubser1: There's a lot of these kinds of stories: 
```
## Story from conversation with cfa8bb9deaf0427498c662745431a282 on December 15th 2018
* get_started_step1
    - action_greet_user
    - slot{"shown_privacy":true}
* chitchat
    - respond_chitchat
* enter_data
    - utter_not_sure
    - utter_possibilities
```

PR #498, opened 2020-05-04
annotations from melinda

erohmensing:
pushing these cause rasa x was yellow, but i think after merging the response selector most of these new intents need to move

### DIFF
--- a/data/nlu/nlu.md
+++ b/data/nlu/nlu.md
@@ -225,6 +225,9 @@
 - yup
 - yyeeeh
 - considering
+- Hmm, not sure what you mean...
+- i am not sure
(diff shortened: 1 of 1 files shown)


melindaloubser1:
There's nothing critical in here, and a lot of it I don't recognize; I'd vote just to close it.

erohmensing:
Actually now that i see the `dialogueflow` annotations I think there might be a lot of stuff from @ArjaanBuijk here, he should take a look

ArjaanBuijk:
I checked it out, and only recognize the dialogueflow item.

I vote to close this one, and I will redo the dialogueflow annotation.

PR #507, opened 2020-05-06
ella and akela annotations

erohmensing:


data/nlu/nlu.md
```diff
@@ -1436,6 +1446,8 @@
 - what languages you prefer more speaking at?
 - which languages do you understand?
 - you know [French](language)
+- we cant converse in french?
+- can we converse in french?
```
erohmensing, line 1450: should we annotate these languages? (There might be others missing here as well)

data/nlu/nlu.md
```diff
@@ -533,6 +538,7 @@
 - when should i use [rasa](product) and when should i use [rasa x](product)
 - what is the difference between [rasa open source](rasa) and [rasa x](product)
 - is [rasa open source](product) needed for [rasa x](product)?
+- what is the difference between rasaand rasa x
```
erohmensing, line 541:
```suggestion
- what is the difference between rasaand [rasa x](product)
```

data/nlu/nlu.md
```diff
@@ -4506,6 +4550,10 @@
 - i dont like bots
 - this is a really frustrating experience
 - you're rather dull
+- Thats so rude
+- sadly
+- That tool here isnt good
+- your crazy
```
erohmensing, line 4556: last one probably an insult

data/nlu/nlu.md
```diff
@@ -1377,6 +1385,8 @@
 - you are ai
 - you are chatbot
 - you're a bot
+- are u human
+- are you human
```
akelad, line 1389: so um i hesitated adding "are you human" to this intent, because the answer to this question is "i am indeed a bot" or similar... maybe we should have a new intent for "are you a human"?

data/nlu/nlu.md
akelad: hm yeah we probably should.. i can see there's a few in this intent that are unannotated.

data/nlu/nlu.md
akelad: yes i don't think i added this one 🤔

data/nlu/nlu.md
```diff
@@ -5088,7 +5152,6 @@
 - rasax
 - RASAX
 - RASA X
--  rasa x
```
akelad, line 5091: um what is this haha

data/nlu/nlu.md
erohmensing: Yeah that's a good point. Was wondering if there was a good response that could cover both, but doesn't seem like it. Then could `are you real` go under `are you human`?

data/nlu/nlu.md
erohmensing: oh i know, i'm also just re-reviewing all the changes in general, i do it when i push my own changes 😄

data/nlu/nlu.md
erohmensing: I think someone must have saved ` rasa x` instead of `rasa x` as a synonym? seems like a weird thing to do so i imagine it must have been annotation from within rasa x 🤔 anyway i do think it should be deleted though

erohmensing: we need to add `x` though https://github.com/RasaHQ/rasa-demo/pull/507/files#diff-d57826d5a178354625db3f820f613afeR3228

data/nlu/nlu.md
erohmensing: removed these 2, making a PR for that here: https://github.com/RasaHQ/rasa-demo/pull/509

Issue #499, opened 2020-05-04
No emoji in weather response 


erohmensing:
![image](https://user-images.githubusercontent.com/22333512/80959087-5ed22600-8e06-11ea-8740-858de7819816.png)

Is this because of response selector?

PR #509, opened 2020-05-06
Separate out "Are you human?" from "are you a bot"?

erohmensing:
grabbed some `human` and `real` nlu data from the instance. still to do: 
- [x] fix what i accidentally put as `ask_is_bot` (it's quite annoying those are still listed)
- [x] add a response in responses.md
- [x] this is me realizing i don't have to write any stories holy fuck i like the response selector

- fixes https://github.com/RasaHQ/rasa-demo/issues/499

### DIFF
--- a/data/nlu/responses.md
+++ b/data/nlu/responses.md
@@ -138,7 +138,7 @@
 
 ## chitchat
 * chitchat/ask_weather 
-  -  I don't know about where you live, but in my world it's always sunny ☀️
+  -  I don't know about where you live, but in my world it's always sunny 🔆
   -  It's getting pretty chilly!
   -  Where I'm from, it's almost never-leaving-the-house weather.
   -  Winter is coming ⚔️
--- a/data/nlu/responses.md
+++ b/data/nlu/responses.md
@@ -72,8 +72,8 @@
 
 ## chitchat
 * chitchat/ask_builder 
-  -  The engineers at Rasa 🛠
   -  One of the smart engineers at Rasa! 🧠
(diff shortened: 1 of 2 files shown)


data/nlu/responses.md
```diff
@@ -72,8 +72,8 @@
 
 ## chitchat
 * chitchat/ask_builder 
-  -  The engineers at Rasa 🛠
   -  One of the smart engineers at Rasa! 🧠
+  -  The engineers at Rasa 🛠
```
erohmensing, line 76: Since people often ask about mother and father, figured the singular one made more sense. Also i miss our varied clever responses 😭

akelad: i am confused, how does having two responses here even work? Or are they just ignored?

erohmensing: All but the first are ignored

erohmensing: It gets quite awkward when users ask for multiple jokes 🙈

erohmensing:
Not terrible! Except for 
- I accidentally fucked up and added stuff to `ask_isbot` because it was still in the list (not technically part of adding a new still)
- There's no way for me to add a response in the UI so i had to come to the command line


Overall i think the workflow of finding something we wanted to handle and searching for similar things in the NLU inbox actually v helpful

data/nlu/responses.md
akelad: ugh

akelad:
let's make sure this positive feedback also gets back to abi/ed? :D

Issue #501, opened 2020-05-05
Tag action server image on merge to master


melindaloubser1:
The action server image should only be tagged and pushed on merge to master

akelad:
@melindaloubser1 can you label this and put it on our board please?

PR #512, opened 2020-05-07
push action server image only on merge

melindaloubser1:
Closes https://github.com/RasaHQ/rasa-demo/issues/501
Action server image will only be pushed on push to master.
Image will be tagged with `run$GITHUB_NUMBER-$GITHUB_RUN_ID` (as well as `latest`)

Reason for the tag change:
Push to master is not guaranteed to be a PR, so we can't always use `pr...`. GITHUB_NUMBER provides an easy way to see what was the latest build since it increments by run, GITHUB_RUN_ID makes it easy to find the run (and therefore commit) that triggered the run since that's what's in a run's URL.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout git repository 🕝
         uses: actions/checkout@v2
 
-        - name: Authenticate into Google Cloud Platform
+      - name: Authenticate into Google Cloud Platform
         uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
         with:
           version: '275.0.0'
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -21,21 +21,21 @@ jobs:
           version: '275.0.0'
           service_account_key: ${{ secrets.GCLOUD_AUTH }}
 
-        - name: Configure Docker to use Google Cloud Platform
-          run: |
-            gcloud auth configure-docker
+      - name: Configure Docker to use Google Cloud Platform
+        run: |
+          gcloud auth configure-docker
 
-        - name: Pull Latest Image
-          run: |
-            docker pull gcr.io/replicated-test/rasa-demo:latest || true
+      - name: Pull Latest Image
+        run: |
+          docker pull gcr.io/replicated-test/rasa-demo:latest || true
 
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,42 @@
+name: Continuous Integration
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - 'actions/**'
+    - 'Dockerfile'
```
erohmensing, line 8: Whoa, this is cool!

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Set Build ID from run ID and number
+        run: echo ::set-env name=BUILD_NUMBER::$GITHUB_RUN_NUMBER-$GITHUB_RUN_ID
+
+      - name: Build Image
+        run: |
+          docker build -t gcr.io/replicated-test/rasa-demo:run$BUILD_NUMBER -t gcr.io/replicated-test/rasa-demo:latest --cache-from gcr.io/replicated-test/rasa-demo:latest .
+      
+      - name: Push PR Image to Google Cloud Container Registry
+        run: |
+          docker push gcr.io/replicated-test/rasa-demo:run$BUILD_NUMBER
```
erohmensing, line 42: How does it also push to latest? Looks like we tag as both but only push to the build number. 

https://console.cloud.google.com/gcr/images/replicated-test/GLOBAL/rasa-demo?project=replicated-test&gcrImageListsize=30 doesn't really list latest

melindaloubser1: yeah I guess it doesn't... I thought by tagging it as both, when it was pushed as one it would also be tagged as the other. Should  we do both?

melindaloubser1: the last `latest` is from jan so no... it's not tagging it as that

melindaloubser1: added latest as well

melindaloubser1: actually! pushing just `rasa-demo` should push both tags. Tested it on helpdesk https://hub.docker.com/r/rasa/helpdesk-assistant/tags

erohmensing: this is really cool, do you think it makes it harder to figure out which tag corresponds to which pr though?

Issue #514, opened 2020-05-07
Remove stopwords and changelog results from docs search


melindaloubser1:
Implement removing stopwords & punctuation before performing docs search
Remove results from change log before returning first two results
see https://github.com/RasaHQ/rasa-demo/issues/485

PR #523, opened 2020-05-18
strip stopwords & punctuation and remove changelog from results

melindaloubser1:
closes https://github.com/RasaHQ/rasa-demo/issues/514

### DIFF
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -685,10 +685,9 @@ class ActionDocsSearch(Action):
             last_user_event = get_last_event_for(tracker, "user", skip=2)
             if last_user_event:
                 search_text = last_user_event.get("text")
-                algolia_result = algolia.search(search_text)	
-        else:	
+                algolia_result = algolia.search(search_text)
+        else:
             algolia_result = algolia.search(search_text)
-
 
         if (
             algolia_result
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -691,9 +691,18 @@ class ActionDocsSearch(Action):
 
         if alg_res and alg_res.get("hits") and len(alg_res.get("hits")) > 0:
             docs_found = True
-            doc_list = algolia.get_algolia_link(alg_res.get("hits"), 0)
+            hits = [
+                hit
+                for hit in alg_res.get("hits")
+                if "Rasa X Changelog " not in hit.get("hierarchy", {}).values()
+                and "Rasa Open Source Change Log "
+                not in hit.get("hierarchy", {}).values()
+            ]
+            if not hits:
+                hits = alg_res.get("hits")
+            doc_list = algolia.get_algolia_link(hits, 0)
             doc_list += (
-                "\n" + algolia.get_algolia_link(alg_res.get("hits"), 1)
+                "\n" + algolia.get_algolia_link(hits, 1)
                 if len(alg_res.get("hits")) > 1
                 else ""
             )
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -675,7 +675,7 @@ class ActionDocsSearch(Action):
         search_text = tracker.latest_message.get("text")
 
         # Search of docs pages
-        alg_res = None
+        algolia_result = None
         algolia = AlgoliaAPI(
             config.algolia_app_id, config.algolia_search_key, config.algolia_docs_index
         )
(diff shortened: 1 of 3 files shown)


actions/actions.py
```diff
@@ -691,9 +691,18 @@ def run(self, dispatcher, tracker, domain):
...
+            hits = [
+                hit
+                for hit in alg_res.get("hits")
+                if "Rasa X Changelog " not in hit.get("hierarchy", {}).values()
+                and "Rasa Open Source Change Log "
+                not in hit.get("hierarchy", {}).values()
+            ]
+            if not hits:
+                hits = alg_res.get("hits")
+            doc_list = algolia.get_algolia_link(hits, 0)
```
melindaloubser1, line 703: Wasn't sure if this belongs here, or in algolia.py.

actions/api/algolia.py
```diff
@@ -20,5 +35,6 @@ def get_algolia_link(self, hits: List, index: int):
         return doc_link
 
     def search(self, search_string: Text):
-        res = self.index.search(search_string)
+        search_text = preprocess_search_text(search_string)
+        res = self.index.search(search_text)
```
melindaloubser1, line 39: same here, not sure if this should be in the class or in the `run` method of the action

actions/actions.py
erohmensing: Hm, i guess it makes sense to keep it here bc it's not algolia specifc, i.e. if we change the names of the rasa changelogs then we have to update it here

Would you mind changing `alg_res` to `algolia_result` here? It took me longer than I should have to interpret what was happening

actions/actions.py
```diff
@@ -691,9 +691,18 @@ def run(self, dispatcher, tracker, domain):
 
         if alg_res and alg_res.get("hits") and len(alg_res.get("hits")) > 0:
             docs_found = True
-            doc_list = algolia.get_algolia_link(alg_res.get("hits"), 0)
+            hits = [
+                hit
+                for hit in alg_res.get("hits")
+                if "Rasa X Changelog " not in hit.get("hierarchy", {}).values()
+                and "Rasa Open Source Change Log "
+                not in hit.get("hierarchy", {}).values()
```
erohmensing, line 699: Would it be easier to check the link url for the word `changelog`?

erohmensing: I think that would have to be after we `get_algolia_link` i guess

actions/actions.py
melindaloubser1: ah sure

actions/actions.py
melindaloubser1: yeah, when I was testing I just checked if any hierarchy level had "change log" or "changelog" in it, but then I thought maybe that could end up funky (and for Rasa X, the compatibility matrix falls under a "Changelog" heading, and that might be a legitimate result sometimes)

actions/actions.py
```diff
@@ -685,16 +685,28 @@ def run(self, dispatcher, tracker, domain):
             last_user_event = get_last_event_for(tracker, "user", skip=2)
             if last_user_event:
                 search_text = last_user_event.get("text")
```
melindaloubser1, line 687:
```suggestion
                search_text = last_user_event.get("text")
                algolia_result = algolia.search(search_text)	
        else:	
            algolia_result = algolia.search(search_text)
```

actions/actions.py
```diff
@@ -685,16 +685,28 @@ def run(self, dispatcher, tracker, domain):
...
-                alg_res = algolia.search(search_text)
-        else:
-            alg_res = algolia.search(search_text)
 
-        if alg_res and alg_res.get("hits") and len(alg_res.get("hits")) > 0:
+        algolia_result = algolia.search(search_text)
```
melindaloubser1, line 689:
```suggestion
```

actions/actions.py
erohmensing: good point!

actions/requirements-actions.txt
```diff
@@ -5,4 +5,5 @@ beautifulsoup4~=4.6.3
 requests~=2.21.0
 geopy~=1.18.1
 pandas~=0.24.1
-algoliasearch>=2.0,<3.0
\ No newline at end of file
+algoliasearch>=2.0,<3.0
+spacy~=2.1.9
```
erohmensing, line 9: bahh i wish we didn't need to have spacy as a dependency it's so large 🙈 ah well

melindaloubser1: We could use a different stopword list if we wanted to get rid of it?

Issue #529, opened 2020-05-22
Fallback flow stuck in a loop on faqs


melindaloubser1:
This conversation happened without ever continuing to final fallback:

```
* faq{"product":"x"}
    - slot{"product":"x"}
    - action_default_fallback
* faq{"product":"x"}
    - slot{"product":"x"}
    - action_default_fallback
* faq{"product":"x"}
    - slot{"product":"x"}
    - action_default_fallback
* faq{"product":"x"}
    - slot{"product":"x"}
    - action_default_fallback
* faq{"product":"x"}
    - slot{"product":"x"}
    - action_default_fallback
```

https://website-demo.rasa.com/conversations/0f8999ab08794347974b86c369af8df3

Issue #519, opened 2020-05-17
no "explain" from action_faq_explain after utter_possibilities


erohmensing:
http://website-demo.rasa.com/conversations/cefef474c37649bcada98bc1c07b2187/1589230149.4084873

After listing the possibilities, if asked to explain, `action_faq_explain` is called, but there is no utterance for the situation.

erohmensing:
It also happens after the open source explanation: http://website-demo.rasa.com/conversations/cefef474c37649bcada98bc1c07b2187/1589230149.4084873

it might be good to have a fallback for the "explain" action -- something like "sorry, that's all the info i've got on the subject"

PR #530, opened 2020-05-22
product entities for FAQs in getting started flow

melindaloubser1:
- add stories for faqs with product entities  as a response to "which product"? & outside of any flow
- don't use entities for chitchat
- added response for "explain" for faqs with no further explanation
- correct setting slot for faqs so that the explain action can access the full retrieval intent name
- fixes https://github.com/RasaHQ/rasa-demo/issues/529
- fixes https://github.com/RasaHQ/rasa-demo/issues/519 

I thought at first we could just do `use_entities: []` for faqs to fix 529, but that breaks the getting started flow if someone answers `rasa open source` or `rasa x`.


data/core/step1_get_started.md
```diff
@@ -476,18 +476,18 @@
...
     - slot{"onboarding": false}
     - utter_ask_which_product
-* faq{"product": "rasa"}
-    - slot{"product": "rasa"}
-    - utter_explain_core
-    - utter_explain_nlu
+* faq{"product": "x"}
+    - slot{"product": "x"}
+    - utter_explain_x
+    - utter_also_explain_nlucore
```
erohmensing, line 491: I think this is the same as the new story you added below -- we should probably keep this one the same

data/core/faqs.md
```diff
@@ -7,6 +7,37 @@
...
+    - slot{"product": "core"}
+    - respond_faq
+    - action_set_faq_slot
+
+## faqs
+* faq{"product": "all"}
+    - slot{"product": "all"}
+    - respond_faq
+    - action_set_faq_slot
+
```
erohmensing, line 39: Re: Your comment about `faq: use_entities: []` -- I _think_ we can actually still have that as long as we have stories where the slot event happens? Because i think use entities only unfeaturizes the entity -- if the categorical slot auto-fills then we can still split on that. wdyt?

actions/actions.py
```diff
@@ -222,7 +224,11 @@ def name(self) -> Text:
...
 
     def run(self, dispatcher, tracker, domain) -> List[EventType]:
-        fullintent = tracker.latest_message["intent"].get("full_retrieval_intent")
+        fullintent = (
+            tracker.latest_message.get("response_selector", {})
+            .get("faq", {})
+            .get("full_retrieval_intent")
+        )
         if fullintent and fullintent.startswith("faq"):
             topic = fullintent.split("/")[1]
```
erohmensing, line 233:
```suggestion
        full_intent = (
            tracker.latest_message.get("response_selector", {})
            .get("faq", {})
            .get("full_retrieval_intent")
        )
        if full_intent and full_intent.startswith("faq"):
            topic = full_intent.split("/")[1]
```
also don't think we need the `startswith` check now that you've changed it to get from `faq`

actions/actions.py
```diff
@@ -209,8 +209,10 @@ def name(self) -> Text:
     def run(self, dispatcher, tracker, domain) -> List[EventType]:
         topic = tracker.get_slot("faq")
 
-        if topic in ["channels", "language", "ee", "slots", "voice"]:
+        if topic in ["channels", "languages", "ee", "slots", "voice"]:
             dispatcher.utter_message(template=f"utter_faq_{topic}_more")
+        else:
+            dispatcher.utter_message(template=f"utter_no_further_info")
```
erohmensing, line 215:
```suggestion
            dispatcher.utter_message(template="utter_no_further_info")
```

actions/actions.py
melindaloubser1: Ah this is true

melindaloubser1:
```suggestion
        full_intent = (
            tracker.latest_message.get("response_selector", {})
            .get("faq", {})
            .get("full_retrieval_intent")
        )
        if full_intent:
            topic = full_intent.split("/")[1]
```

data/core/faqs.md
melindaloubser1: So, I thought that this is what happens, but experimentation suggested it wasn't. I'd like to test it out again though because I might've missed something (and I can't remember exactly what I tried now of course =P)

data/core/step1_get_started.md
```diff
@@ -476,18 +476,18 @@
...
```
melindaloubser1, line 490:
```suggestion
* faq{"product": "rasa"}
    - slot{"product": "rasa"}
    - utter_explain_core
    - utter_explain_nlu
```

data/nlu/nlu.md
```diff
@@ -626,7 +626,7 @@
 - can someone help me with infos about the enterprise edition
 - can you tell me about rasa x ee?
 - enterprise features
-- i want to learn more about Rasa X EE
+- i want to learn more about [Rasa X](product) EE
```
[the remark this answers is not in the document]
melindaloubser1: Yeah this seems like a mistake
```suggestion
- i want to learn more about Rasa X EE
```

data/nlu/nlu.md
```diff
@@ -5379,7 +5379,7 @@
 ## synonym:x
 - Rasax
 - RASAX
-- Rasa X
+- [Rasa X](product)
```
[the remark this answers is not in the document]
melindaloubser1: you're quite right
```suggestion
- Rasa X
```

data/core/faqs.md
melindaloubser1: you mean using this, with `use_entities=[]` for faqs should work right?
```
## faqs
* faq
    - slot{"product": "x"}
    - respond_faq
    - action_set_faq_slot

## new to rasa at start
* how_to_get_started{"user_type": "new"}
    - action_set_onboarding
    - slot{"onboarding": true}
    - utter_getstarted_new
    - utter_built_bot_before
* deny
    - utter_explain_rasa_components
    - utter_rasa_components_details
    - utter_ask_explain_nlucorex
* faq
    - slot{"product":"rasa"}
    - utter_explain_nlu
    - utter_explain_core
```

melindaloubser1: Ok, I just tested it again, when `use_entities=[]` then the slot is not autofilled. Is this behaviour documented?

melindaloubser1:
@erohmensing I merged it but the question you asked above is still valid - might be either undocumented or undesirable behaviour

PR #533, opened 2020-05-26
add restart intent data

### LINKED ISSUE #528: Add nlu data for /restart intent
Sometimes people ask to restart the convo without using `/restart`, adding some nlu data would make that possible
https://website-demo.rasa.com/conversations/01eb53997a9b4d1b991939c3f74ee0cb/1588684745.222202

erohmensing:
close #528

An open question about this -- from looking at the NLU data, it seems pretty common (even maybe more common) for users to ask things like "how to restart rasa server" or questions about restarting rasa bots in general. In that case, it would be pretty unintuitive to get a "congrats, you restarted me!" -- perhaps it'd be nicer if we gave them the "click the button below to restart" option instead. we've have to hack it by creating a new intent not titled exactly `restart` though, i think. not sure if we'd want to do that? 🤔


melindaloubser1:
I think it'd be good to add that other kind of restart data to technical questions; that might be good enough that they don't get restarted accidentally too often.

erohmensing:
Good point! I'll add data from that and run a full nlu test and see if they're often confused

erohmensing:
@me realizing you did the exact same thing already 🙃

thank you, and great minds think alike

melindaloubser1:
hmm it didn't post the CV results though. I'll see what's up with that.

melindaloubser1:
looks like nlu testing just didn't run. Maybe there's a mistake in the condition? I dunno.

PR #543, opened 2020-06-02
Rasa X annotations

b-quachtran:


b-quachtran:
Rasa X is doing some weird things with the NLU files, going to have to undo some of the deletes it's making

akelad:
@melindaloubser1 do you know why github actions no longer posts nlu eval results?

melindaloubser1:
@akelad Since we changed it to only run cv if there's a `nlu_testing_required` tag present, it just didn't run the eval at all.

PR #555, opened 2020-07-06
Download spacy en

ArjaanBuijk:
Without downloading the spacy language model, the container crashes here during startup:

```
  File "/app/actions/actions.py", line 21, in <module>
    from actions.api.algolia import AlgoliaAPI
  File "/app/actions/api/algolia.py", line 7, in <module>
    en_spacy = spacy.load("en")
  File "/opt/venv/lib/python3.7/site-packages/spacy/__init__.py", line 27, in load
    return util.load_model(name, **overrides)
  File "/opt/venv/lib/python3.7/site-packages/spacy/util.py", line 139, in load_model
    raise IOError(Errors.E050.format(name=name))
OSError: [E050] Can't find model 'en'. It doesn't seem to be a shortcut link, a Python package or a valid path to a data directory.
```


Dockerfile
```diff
@@ -20,6 +20,10 @@ COPY ./actions /app/actions
 COPY setup.py /app
 RUN  pip install -e . --no-cache-dir
 
+# Download spacy language data
+RUN python -m spacy download en_core_web_md
+RUN python -m spacy link en_core_web_md en
+
```
erohmensing, line 26:
```suggestion
# Download spacy language data
RUN python -m spacy download en_core_web_sm
RUN python -m spacy link en_core_web_sm en

```
I think this should be fine if we're just using it for stopwords!

PR #556, opened 2020-07-07
Rasa x jul 7th 20  3 20 01 pm ben

b-quachtran:


melindaloubser1:
Just one question, why does it look like the lookups table changed?

b-quachtran:
> Just one question, why does it look like the lookups table changed?

I think it's Rasa X causing that phantom diff. I've seen it happen before when it tries to re-format the NLU files.

Issue #567, opened 2020-08-27
chitchat intents missing from intent description mapping 


erohmensing:
https://github.com/RasaHQ/rasa-demo/blob/master/actions/intent_description_mapping.csv
When we ported chitchat to retrieval intents, we didn't add them to the intent description mapping!

PR #568, opened 2020-08-27
handle chitchat intents in intent description mapping

erohmensing:
close https://github.com/RasaHQ/rasa-demo/issues/567


melindaloubser1 (approved):
Ooh thanks fro catching this!

PR #571, opened 2020-09-01
Rasa X annotations

TyDunn:


melindaloubser1:
Looks like this was a single annotation to `faq` instead of a subintent

PR #583, opened 2020-10-06
Add need_help_broad & utter_ask_more_details

samsucik:
I added the new `need_help_broad` intent which should cover messages where the user indicates they need help or want to ask a question, but they should be more specific. I also annotated several examples. Some selected ones:
```
i need help regarding chatbot dev
need help on chatbot
I have a question
I am stuck and I need help
I have a few questions
there is an issue during installation
```

Further, I added a preferred response to the above messages -- `utter_ask_more_details` -- together with a few response templates.

There's now also a very simplistic rule taking care of this all.


akelad:
requested ben because he has a better overview of this atm

b-quachtran (changes requested):
Looks good! Could you also add two rules so that the newsletter & sales forms get deactivated when this intent is triggered?

samsucik:
@b-quachtran thanks! I've added the rules so that `need_help_broad` deactivates the loop. Though if we later use this intent also to cover questions related to the active form, we might want to revisit this then. For now, let's go with this simple version and hope that `action_explain_sales_form` covers most of the questions people have about the form :-)

b-quachtran (approved):
🚀

PR #590, opened 2020-10-21
Make pronouns general where they have no reason to be specific.

samsucik:


erohmensing (approved):
Thank you! 

Would you call a chatbot a "they" or an "it"? I'm not sure 😄

samsucik:
I'd go with the animate version since we refer to Sara as "she/her" but good question! And yeah, I agree, let's remove that one response variation.

PR #598, opened 2020-12-03
Rasa X Annotations - NLU Insights

b-quachtran:


b-quachtran:
@dakshvar22 Here's what the CLI tool provided as far as recommendations: 
Screen Shot 2020-12-03 at 9 09 34 AM

dakshvar22:
Thanks @b-quachtran That's exactly what I thought. The `MinimumExampleLinter` is actually a noisy linter if the F1 score is not taken into consideration. When comparing the results with the results on this [PR](https://github.com/RasaHQ/rasa-demo/pull/566) - `next_step`, `pipeline_recommendation` and `nlu_generation_tool_recommendation` have really small number of training examples but the F1 score is still quite high and hence they should actually be neglected from the suggestions.

@wochinge We should change the `MinimumExampleLinter` accordingly.

dakshvar22:
@b-quachtran I am just wondering why the warnings here are not the same as the warnings that @wochinge got [here](https://github.com/RasaHQ/rasa-x/pull/3891#issuecomment-721659359). I was expecting other linters to suggest more intents. Could be two reasons - 

1. The underlying training data is different in both the cases.
2. The CLI tool was run with different parameters.

b-quachtran:
@dakshvar22 Hmm, I ran it using the local setup on the latest master data. I think the difference is with the Remote setup, so I might need to re-do this trial

b-quachtran:
@dakshvar22 Yeah the parameters were different. I'll re-do this insights run today and tag you in the new PR

Issue #535, opened 2020-05-26
add actions unit tests


erohmensing:
we should be unit testing our actions, especially ones with external apis (like newsletter and sales, as long as there is a way to then delete our entries 🤔) and the event one which reads our website's HTML (we ran into bugs with this one before when the website layout changed)

PR #607, opened 2021-02-04
Unit tests

melindaloubser1:
- Create unit tests for custom actions, focusing on external APIs
- Closes https://github.com/RasaHQ/rasa-demo/issues/535
- Rename workflows more intuitively and add action unit testing to CI pipeline 
    - remove mergepal  - use automerge instead
    - add running unit tests
    - rename workflows & files for what they actually do


melindaloubser1:
@erohmensing not all actions are covered by these, but since it's getting to be a lot of changes already, I think we should get this mergeable and in before adding more. Wdyt?

erohmensing:
I agree, it makes sense to do it in phases, i'll take a look!

.pre-commit-config.yaml
```diff
@@ -3,4 +3,3 @@ repos:
     rev: stable
     hooks:
     - id: black
-      language_version: python3.6
```
melindaloubser1, line 6: specifying version causes hook not to work if it's not a python 3.6 env

.github/workflows/continuous_integration.yml
```diff
@@ -1,42 +1,191 @@
...
+          ( 
+            ( 
+              contains( github.event.pull_request.labels.*.name, 'nlu_testing_required' ) 
+              && contains(  steps.files.outputs.all, 'data/nlu/' ) 
+            )
+            || contains(  steps.files.outputs.all, 'config.yml' ) 
+          )
+      run: |  
+          rasa test nlu -f 3 --cross-validation
+          python .github/workflows/format_results.py
```
melindaloubser1, line 129: It seems to be ignoring this condition now, it ran cv although there was no label 🤔

melindaloubser1:
@erohmensing lmk if I should rather separate out the CI/CD workflow changes in a separate PR. I got on a roll and it got a little out of hand 😬

tests/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+from actions import config
+from actions.api.rasaxapi import RasaXAPI
+from actions.api.mailchimp import MailChimpAPI
+from actions.api.gdrive_service import GDriveService
+
+here = Path(__file__).parent.resolve()
+
+@pytest.fixture
+def tracker():
+    tracker = Tracker.from_dict(json.load(open(here / "./data/initial_tracker.json")))
```
erohmensing, line 20:
```suggestion
    with open("tests/data/initial_tracker.json") as json_file:
        tracker = Tracker.from_dict(json.load(json_file))
```

I think the relative path here works, you definitely want to use `with open()` anytime you read a file, because it will autoclose after you're done

tests/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+    yield email
+    client.unsubscribe_user(config.mailchimp_list, email)
+
+@pytest.fixture
+def mailchimp_subscribed_email():
+    """make sure the fake email is subscribed"""
+    email = "example_subscribed@rasa.com"
+    client = MailChimpAPI(config.mailchimp_api_key)
+    client.subscribe_user(config.mailchimp_list, email)
+    yield email
```
erohmensing, line 71: is this one going to get an error when we try to set it up again, if we don't tear it down (unsubscribe the user) after yield?

tests/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+@pytest.fixture
+def rasa_x_conversation_endpoint():
+    return f"{RasaXAPI.schema}://{RasaXAPI.host}/api/conversations"
+
+@pytest.fixture
+def rasa_x_auth_header():
+    return RasaXAPI.get_auth_header()
+
+@pytest.fixture
+def rasa_x_convo(rasa_x_conversation_endpoint, rasa_x_auth_header, tracker):
```
erohmensing, line 44: Would be easier to understand if these were described as what the fixtures do , sth like "Create a conversation for test to add a tag to" and "Create a user who is already subscribed to the newsletter". same with the names of the mailchimp fixtures, I think

actions/api/community_events.py
```diff
@@ -54,53 +59,54 @@ def as_kwargs(self) -> Dict[Text, Text]:
...
+            logger.warning(e)
+            return (
+                datetime.datetime.max
+            )  # if date can't be parsed assume event is future
 
-    if response.status_code == 200:
-        community_page = response.content
+    @staticmethod
+    def get_community_page() -> requests.Response:
+        return requests.get("https://rasa.com/community/join/")
```
erohmensing, line 77: makes sense to pull this out, can we just make this a constant at the top of the module so we know where to find it if it changes?

erohmensing (changes requested):
Great job with these! The fixtures are super thorough, i appreciate you giving time and thought to how to make these rerunnable without filling up our instance/newsletter with dummy conversations and such 😄
A few overall comments: 
- I think it makes sense to keep the functions for getting community events outside of the community event class (still under the same API though, where they currently are), as they aren't particular to specific event instances
- Generally, consistency in the docstrings (uppercase vs lowercase, how the quotes are positioned

Coming back to review the tests themselves tomorrow -- mind blacking them in the meantime? Im so used to it now its faster to read 🙈

Makefile
```diff
@@ -26,4 +45,7 @@ lint:
 	black --check actions
 
 types:
-	pytype --keep-going actions
\ No newline at end of file
+	pytype --keep-going actions
```
erohmensing, line 48: On that note ^ 
`pytype --keep-going actions tests`

as well as `black actions tests` and `flake8 actions tests`

.github/workflows/continuous_integration.yml
```diff
@@ -53,6 +53,37 @@ jobs:
...
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        make install-dev
+    - name: Unit Tests
+      working-directory: ${{ github.workspace }}
+      run: |
+        make test-actions
+    - name: Install dependencies
```
melindaloubser1, line 86:
```suggestion
```

Makefile
```diff
@@ -8,6 +12,21 @@ help:
...
+	@echo "        Run custom action unit tests"
+
+install:
+	pip install -r requirements.txt
+	python -m spacy download en_core_web_md
+	python -m spacy link en_core_web_md en 
+	pip install -e .
+
+install-dev:
+	python -m pip install --upgrade "pip<20"
```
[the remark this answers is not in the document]
melindaloubser1: Ope I was going off of what we had in our CI. I guess we should either put this in install and install-dev or in neither. The pip<20 is there (I believe) to avoid issues with pip's new dependency resolver, which can take forever to install stuff

Makefile
melindaloubser1, line 48:
```suggestion
	pytype --keep-going actions tests
```

actions/api/mailchimp.py
```diff
@@ -44,6 +52,22 @@ def subscribe_user(self, list_id: Text, email: Text) -> bool:
...
+                    # if user is in database, but is unsubscribed or archived, attempt to resubscribe
+                    self.client.lists.members.update(
+                        list_id,
+                        hash,
+                        data={"email_address": email, "status": "pending"},
+                    )
+                    return True
+                return False
+            except MailChimpError:
+                return False
```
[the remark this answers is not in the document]
melindaloubser1: We'll hit L67 when  user is subscribed or pending. We'll only hit L69 if something else goes wrong - e.g. MailChimp can't handle the email address, or fails to make an update because of a temporary interruption, something like that. 

Ja, I'm in favour of switching the order of conditions - we're already not relying on exceptions except for basically unknown mailchimp issues

actions/api/rasaxapi.py
```diff
@@ -0,0 +1,47 @@
...
+logger = logging.getLogger(__name__)
+
+
+class RasaXAPI:
+    """Class to connect to the Rasa X API"""
+
+    schema = config.rasa_x_host_schema
+    host = config.rasa_x_host
+    username = config.rasa_x_username
+    password = config.rasa_x_password
```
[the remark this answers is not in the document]
melindaloubser1: I wanted them to be class attributes (not instance), which is why they are here, but I'm not clear on which is the right way - is it better to force using an instance than a class?

actions/requirements-actions.txt
```diff
@@ -6,4 +6,5 @@ requests~=2.24.0
 geopy~=1.18.1
 pandas~=0.24.1
 algoliasearch>=2.0,<3.0
-spacy~=2.2.4
\ No newline at end of file
+spacy~=2.2.4
+pytest-mock~=3.5.1
```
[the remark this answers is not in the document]
melindaloubser1: dev, you're right, action server won't need this

tests/conftest.py
melindaloubser1: No, it needs to be subscribed anyway, the fixture just ensures it is. However it would be better to tear it down so we don't leave random fake users in the mailchimp list.

actions/api/mailchimp.py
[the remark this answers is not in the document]
melindaloubser1: Actually wait! We'll also hit L69 when the user isn't in the database at all, because in that case retrieving the status will fail, so it will go to except. This seems faulty. Will look into it.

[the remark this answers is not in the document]
melindaloubser1: Ok no, the first try block takes care of that. Added more explanatory comments. I think it makes sense to rely on try/excepts here, either first to try user creation or fetching user status, but we could further change the order of attmpts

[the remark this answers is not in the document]
melindaloubser1: Sorry about that. Can you take a look at the updated function and how it's called in actions now?

actions/api/rasaxapi.py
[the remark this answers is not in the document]
erohmensing: I guess the upside of instances would be that if we at some point wanted to connect to two different rasa x apis or as two different users or something, we could do that, but not sure that's actually necessary 😄

erohmensing:
A few comments got lost in the accodion btw -- about the domain getting reformatted

actions/api/mailchimp.py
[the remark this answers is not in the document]
erohmensing: Yes, it looks great! The comments are super helpful. If it makes it easier we could always just return a string from `["user_subscribed", "already_subscribed", "failure"]` back to the actions to interpret -- but I think your comments are clear enough as well

tests/test_actions.py
```diff
@@ -0,0 +1,217 @@
...
+    name = uuid.uuid1().hex
+    job_function = "Test Actions"
+    email = "example@rasa.com"
+
+    tracker.slots["budget"] = budget
+    tracker.slots["company"] = company
+    tracker.slots["business_email"] = email
+    tracker.slots["job_function"] = job_function
+    tracker.slots["person_name"] = name
+    tracker.slots["use_case"] = use_case
```
erohmensing, line 204: this could be simplified with a 
```
collected_info = {
... dict of slots
}
tracker.slots.update(collected_info)
```
doesn't make it any shorter, but i think its the better pythonic way to do it

erohmensing (approved):
I'm stoked about these! but please also check the comments about the domain, I think it will break the forms as it is currently

actions/api/mailchimp.py
```diff
@@ -35,15 +42,38 @@ def is_valid_email(email: Text) -> bool:
...
 
-        # if the user is already subscribed, return False
         except MailChimpError:
-            return False
+            # user is already in the database, or new user creation failed for another reason
+            try:
+                hash = self.hash_email(email)
+                # check status of user already in database. Will throw exception if they are not in database.
+                status = self.client.lists.members.get(list_id, hash).get("status")
+                if status != "subscribed" or status != "pending":
```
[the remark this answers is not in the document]
melindaloubser1: `not in` but yes

tests/test_actions.py
```diff
@@ -0,0 +1,217 @@
...
+def test_get_discourse_links():
+    """Test that the link returned is correctly formatted and goes to a valid webpage"""
+    # This could also stop at discourse.query and assert that the return matches the mocked API used for other tests
+    discourse = DiscourseAPI("https://forum.rasa.com/search")
+    disc_res = discourse.query("rasa")
+    disc_res = disc_res.json()
+    link_string = discourse.get_discourse_links(disc_res.get("topics"), 0)
+    link = link_string.split("](")[1][:-1]
+    link_result = requests.get(link)
+    assert link_result.status_code == 200
```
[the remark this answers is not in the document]
melindaloubser1: I've split it up and added some. Anything that runs an action is in test_actions, anything that directly calls an api method is in that api's file

actions/api/mailchimp.py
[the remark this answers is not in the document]
melindaloubser1: Oh lol no you were right and I spent quite a few min figuring out why no one ever got subscribed

actions/api/rasaxapi.py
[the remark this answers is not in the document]
melindaloubser1: Ah, that is fair, i'll move them to be instance

melindaloubser1:
@erohmensing I know you already approved, but restructuring the tests to match the actions folder changed enough things that I think it needs another look

actions/api/community_events.py
```diff
@@ -44,6 +45,11 @@ def from_html(cls, html) -> Optional["CommunityEvent"]:
             link.strip(),
         )
 
+    def __repr__(self) -> Text:
+        return "{} ({}): {} ({})".format(
+            self.name, self.city, self.formatted_date, self.date
+        )
+
     def name_as_link(self) -> Text:
         return "[{}]({})".format(self.name, self.link)
```
erohmensing, line 54: These might be clearer with f strings

actions/api/mailchimp.py
```diff
@@ -35,15 +42,48 @@ def is_valid_email(email: Text) -> bool:
...
+                    # if user is in database, but is unsubscribed or archived, attempt to resubscribe
+                    self.client.lists.members.update(
+                        list_id,
+                        hash,
+                        data={"email_address": email, "status": "pending"},
+                    )
+                    return "newly_subscribed"
+
+            except MailChimpError:
+                # if the subscription of an unsubscribed user fails, return `None` instead of `False` to indicate failure
```
erohmensing, line 74: can update this now (looks so much clearer in the action!)

actions/config.py
```diff
@@ -1,45 +1,34 @@
...
 
 algolia_search_key = os.environ.get("ALGOLIA_SEARCH_KEY", "")
 
 algolia_docs_index = os.environ.get("ALGOLIA_DOCS_INDEX", "")
 
 rasa_x_host = os.environ.get("RASA_X_HOST", "rasa-x:5002")
+
+rasa_x_password = os.environ.get("RASA_X_PASSWORD", "")
+
+rasa_x_username = os.environ.get("RASA_X_USERNAME", "")
```
erohmensing, line 32: should we default to `me` since a lot of people will likely be using local mode?

domain.yml
```diff
@@ -1,51 +1,93 @@
...
-- affirm
-- ask_how_contribute
-- ask_question_in_forum
-- ask_when_next_event
-- ask_which_events
-- ask_why_contribute
-- bye
-- canthelp
+- affirm:
+    use_entities: true
```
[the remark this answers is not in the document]
melindaloubser1: Umm I did not do this on purpose

tests/api/test_algolia.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_get_algolia_link():
+    """Test that the link returned is correctly formatted and goes to a valid webpage"""
+    algolia = AlgoliaAPI(
+        config.algolia_app_id, config.algolia_search_key, config.algolia_docs_index
+    )
+    algolia_result = algolia.search("rasa")
+    link_string = algolia.get_algolia_link(algolia_result.get("hits"), 0)
+    link = link_string.split("](")[1][:-1]
```
erohmensing, line 14: > Test that the link returned is correctly formatted

Maybe add an assertion of a regex match before we go about splitting the link string? That'd help to catch any regression without the test failing in a weird way when you try to split it

tests/api/test_community_events.py
```diff
@@ -0,0 +1,38 @@
+import datetime
+import pickle
+import logging
+
+from actions.api import community_events
+
+
+def test_get_community_events(caplog):
```
erohmensing, line 8: Add types to any arguments (no return type necessary)

tests/api/test_community_events.py
```diff
@@ -0,0 +1,38 @@
...
+        community_events.get_community_events()
+    assert "Error when trying to parse event details from html." not in caplog.text
+
+
+def test_parse_community_events(mocker):
+    mocker.patch.object(
+        community_events,
+        "get_community_page",
+        return_value=pickle.load(open("tests/data/events_page.pkl", "rb")),
+    )
```
erohmensing, line 22: To be safe, I think we want to do this in a `with open`, so something like
```suggestion
    with open("tests/data/events_page.pkl", "rb") as events_page: 
        mocker.patch.object(
            community_events,
            "get_community_page",
            return_value=pickle.load(events_page),
        )
```

tests/api/test_discourse.py
```diff
@@ -0,0 +1,14 @@
...
+
+def test_get_discourse_links():
+    """Test that the link returned is correctly formatted and goes to a valid webpage"""
+    discourse = DiscourseAPI("https://forum.rasa.com/search")
+    disc_res = discourse.query("rasa")
+    disc_res = disc_res.json()
+    link_string = discourse.get_discourse_links(disc_res.get("topics"), 0) or ""
+    link = link_string.split("](")[1][:-1]
+    link_result = requests.get(link)
+    assert link_result.status_code == 200
```
erohmensing, line 14: This will fail if the link_string is `""` right?

tests/conftest.py
```diff
@@ -0,0 +1,162 @@
...
+import json
+import requests
+import uuid
+import sqlalchemy as sa
+from sqlalchemy.orm import sessionmaker
+from mailchimp3.mailchimpclient import MailChimpError
+
+from rasa_sdk.executor import CollectingDispatcher
+from rasa_sdk import Tracker
+from rasa.shared.core.domain import Domain
```
erohmensing, line 23: I think we want to use the rasa_sdk `Domain` here, which is already a dict

tests/conftest.py
```diff
@@ -0,0 +1,162 @@
...
+
+    client.delete_user(config.mailchimp_list, email)
+
+
+@pytest.fixture
+def mailchimp_new_email(mailchimp):
+    """Create a user who is not already subscribed to the newsletter"""
+    email = mailchimp[0]
+
+    yield email
```
erohmensing, line 117: any of these that don't have teardown code can just `return`, i think

tests/conftest.py
```diff
@@ -0,0 +1,162 @@
...
+    # use a random email to avoid mailchimp errors due to too many attempted signups by the same email address
+    email = f"{uuid.uuid4().hex}@rasa.com"
+    client = MailChimpAPI(config.mailchimp_api_key)
+    # try to delete user in case of an improperly exited test
+    try:
+        client.delete_user(config.mailchimp_list, email)
+    except MailChimpError:
+        pass
+
+    yield email, client
```
erohmensing, line 107: This confuses me when it gets used -- can we make 2 separate fixtures for the client and the mailchimp user?

actions/config.py
melindaloubser1: yeah that is fair

tests/api/test_discourse.py
melindaloubser1: Yes, this was a shortcut to avoid error when the return is `None`

tests/conftest.py
melindaloubser1: It is, but we want to load the actual domain of the bot - I started using an empty dict, but realized that could mask problems in the case that an action starts looking in the domain.

PR #613, opened 2021-02-26
Add NLU test examples for intents currently not covered

samsucik:
There are several intents that are covered in the training examples but not in the test examples. I've tried to correct this (by adding real examples from the NLU Inbox) as part of my efforts around updating the Sara dataset in our `training-data` repo.


b-quachtran (approved):
LGTM! The failed CI checks look like a result from some NumPy version compatibility issues and I'll take care of this in a separate PR.

PR #616, opened 2021-03-01
Add NLU training examples for  chitchat/ask_whoisit

samsucik:
These intents weren't previously covered in the training data -- but we clearly used them in the past, so maybe the training data just disappeared at some point? In any case, I now added a few real examples from the NLU inbox because I want to consolidate the dataset a bit for the purposes of NLU regression testing.


erohmensing (approved):
Thank you! 🙏🏻

PR #617, opened 2021-03-01
Update NLU test data

samsucik:
The test data is out-dated:
- chitchat examples are still listed as separate top-level intents
- `ask_when_next_event` is still in there even though it isn't used anymore

I fixed both issues. Note that `ask_when_next_event` still needs to be removed from the domain and from the training stories.


samsucik:
Oh, no, I pushed some changes from the other PR 🤦  Gonna revert that.

tests/test_nlu.yml
```diff
@@ -233,11 +233,7 @@ nlu:
     - what are you good at?
     - options
     - cool! can I do something else here?
-- intent: ask_when_next_event
-  examples: |
-    - what's the next rasa event
-    - is there a Rasa event in [San Francisco](location)
```
erohmensing, line 239: Maybe we should add these examples to the other intent where they live now? :)

erohmensing (approved):
Thanks, would you please create an issue for the outstanding to-do items you noticed with regards to this intent?

tests/test_nlu.yml
samsucik: Oh, totally. Adding it now 🙂

samsucik:
> would you please create an issue for the outstanding to-do items

Done 🙂

PR #619, opened 2021-03-06
Move tagging to RasaXAPI class

melindaloubser1:
- Create a  class for connecting to Rasa X and use this for tagging conversations.
- Remove unused config variables

This came up during creating unit tests -  the Rasa X API is the only one we connect to without a dedicated class, and it is more extendable in a class than in separate methods.


melindaloubser1:
I ended up incorporating this in the unit tests anyway because it made writing the tests a lot easier

PR #620, opened 2021-03-10
Story fragments -> rules

samsucik:
These small changes have been lingering in my code editor and I just wanted to push them before making some other (unrelated) changes to the training data.

I've turned 2 story fragments into rules but there are many, many more stories that should be changed like this.

### DIFF
--- a/data/rules/rules.yml
+++ b/data/rules/rules.yml
@@ -58,18 +58,6 @@ rules:
   steps:
   - intent: next_step
   - action: action_next_step
-- rule: faq (when product slot gets set)
-  steps:
-  - intent: faq
-  - slot_was_set:
-    - product # the slot value doesn't matter
-  - action: utter_faq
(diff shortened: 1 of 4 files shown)


samsucik:
@melindaloubser1 fyi I've removed the `faq -> utter_faq` rule I had introduced. It was creating contradictions because sometimes we mis-use `intent:faq` for utterances like "rasa x" (the same way we mis-use the `how_to_get_started` and `enter_data` intents for this) and then the next action isn't `utter_faq`. And I couldn't be bothered to go down this deep rabbit hole and fix all of that.

melindaloubser1:
haha that is fair. Let's merge this, and if you think it's worth it could you create an issue for the outstanding rabbit hole?

samsucik:
Yep, the issue is now [here](https://github.com/RasaHQ/rasa-demo/issues/623).

PR #622, opened 2021-03-11
workflow changes

melindaloubser1:
- Consolidate workflows into a continuous integration flow that runs on PRs, and a continuous deployment flow that runs on merge.
- Use automerge instead of mergepal
- Use RasaHQ changed files filter 
- Use job level conditions (available since mid-2020) instead of step-level conditions


melindaloubser1:
@HotThoughts Do you know what is necessary here to enable auto-update on auto-merge? I saw in the rasa repo that this workflow https://github.com/RasaHQ/rasa/blob/main/.github/workflows/automatic-pr-update.yml still exists although we're now using Github's automerge, and wasn't sure if I also needed one on this repo

HotThoughts:
> @HotThoughts Do you know what is necessary here to enable auto-update on auto-merge? I saw in the rasa repo that this workflow https://github.com/RasaHQ/rasa/blob/main/.github/workflows/automatic-pr-update.yml still exists although we're now using Github's automerge, and wasn't sure if I also needed one on this repo

I guess we need to enable both since they have different duties. GitHub's auto-merge is to merge the PR to the base branch if the PR is reviewed and all checks passed, while auto-update is to merge base branch to the PR when base branch is updated. 

In short, they both deal with the PR and the base branch, but have opposite directions. 
- Auto-merge: PR -> main
- Auto-update: main -> PR

> If you enable auto-merge for a pull request, the pull request will merge automatically when all required reviews are met and status checks have passed. Auto-merge prevents you from waiting around for requirements to be met, so you can move on to other tasks.

melindaloubser1:
Ah! That makes sense. So I should use a similar workflow to the one in the rasa repo here then. Do you know where I get `RASABOT_AUTOMERGE_GITHUB_TOKEN`?

HotThoughts:
🤔 We need someone who has the access to the rasabot account and also can add the token to this repo. I guess @tmbo ?

melindaloubser1:
Right, if I generate a personal access token that  wouldn't be tied to the org but to my account, so I think an org admin needs to do it.

tmbo:
I've added the secret.

melindaloubser1:
Thank you! @HotThoughts it's ready for review then.

HotThoughts (approved):

HotThoughts:
LGTM 🥇

PR #624, opened 2021-03-12
Better support for the explain intent

samsucik:
We previously had the `explain` intent but it wasn't widely supported -- often, Sara wouldn't know how to react. I've added some new rules and stories to improve this, especially by triggering `utter_no_further_info` where there's no good explanation available today.

My changes don't fix everything, but I tried to be as exhaustive as possible. I went through the training stories as well as through [the new test set](https://github.com/RasaHQ/training-data/pull/9) (collected from real convos) and considered every occurrence of `intent:explain`. In most places, I merely factored out shared patterns from existing training stories into rules.

**Importantly, I couldn't make some rules into actual rules because of a bug in rules (see the actual changes for details). We can wait for a fix, but I'd personally proceed with merging this in now, and later converting the rule-like stories into rules.**

Also importantly: I didn't make any changes to the wording of `utter_no_further_info`. I do think we should improve it to be (and sound) more friendly & helpful now that it's gonna me used more widely. However, this wasn't the aim of my changes. I merely wanted to address this issue **on the action level.**

I tested my changes by training a model with the same random seed and testing against the new test set -- for both this branch and for `main`. The results didn't show any significant regressions -- on the contrary, more actions actually improved. This is on the `good` stories from the new test set (the F1 used to calculate F1 differences, and the support, are both taken from the story reports):
Screenshot 2021-03-12 at 11 14 37
And this is on the (much more interesting, also because they contain many instances of `explain`) `bad` stories:
Screenshot 2021-03-12 at 11 17 41

I'm happy to provide the actual story reports and failed stories too, though I myself also did manual inspection of the utterances which became slightly worse with my changes and didn't find anything serious (as in, no _new_ issue).


data/stories/chitchat.yml
```diff
@@ -732,9 +734,13 @@ stories:
...
-  - intent: nlu_info
-    entities:
-    - nlu_part: intent classification
+  - or:
+    - intent: nlu_info
+      entities:
+      - nlu_part: intent classification
+    - intent: explain
+      entities:
+      - nlu_part: intent classification
```
b-quachtran, line 743: Do the NLU examples need to be updated for the explain intent to extract the `intent classification` and `entity recognition` entities?

b-quachtran (changes requested):
Thanks for getting this update in! The new rules and stories look good to me, and agreed on merging it in as-is for now until that rule issue with actions first is resolved.

Do you think you could also update the conversation test cases for Sara with a few example stories using the `explain` intent?

data/stories/chitchat.yml
samsucik: Good question. However sensible it is, I might actually disagree: Looking at the NLU inbox, I can't find real examples classified as `explain` that would have those entities. On the other hand, there's a lot of real examples classified as `nlu_info` which do sound close to `intent:explain` too, e.g.:
```
No explain entity recognition
can you explain duckling to me
explain intents
explain how to create entity file
```

Assuming we really want to keep classifying these as `nlu_info`, I wouldn't move any of them in the training data for `explain`. Then, I'd treat the relevant training story as a safety net for cases where NLU predicts `explain` -- so that Core still does the right thing then.

After all, my adding of `explain` in places where `nlu_info` is in stories was inspired by several existing training stories where there was an `or` between `nlu_info` and `explain`. In other words, this adding wasn't (in some cases) inspired by any real conversations, but rather by this "safety net" pattern.

samsucik:
> Do you think you could also update the conversation test cases for Sara with a few example stories using the explain intent?

@b-quachtran I'm tempted to say no. So, there's actually a lot of relevant stories in the new test set, but I didn't want to add any of the test set to this repo because:
- it'd take many stories to actually achieve decent coverage of the changes I've made (and there are some changes that were made as "safety net" ones and there aren't necessarily good real convos to test them yet)
- the current test set in `rasa-demo`, I believe, is tiny, low-coverage and low-quality, with out-dated bits, and hence I personally don't have the motivation to maintain it (this was also the main reason why I curated a new big test set). If anything, I think we might want to scrap the old test set and replace it with an end-to-end one from real conversations once Sara is changed for the new docs? 🤔 
- the new test set contains full conversations and it hasn't been anonymised so far; I think it's currently only for internal purposes (though one could remove all the actual text from a few stories)

If you think it really makes sense to still get a small number of stories in, I can do that, but I'm hesitant as I explained 🙂

data/stories/chitchat.yml
b-quachtran: I think it'd make sense to move `nlu_info` into `explain` since there'd be quite a bit of ambiguity otherwise. The `nlu_part` entity would have to be renamed to something more general but IMO it'd be a much cleaner way of handling all these cases.

b-quachtran:
@samsucik Are there any plans to merge some version of the new test set into `rasa-demo`? If there are, I see no issues with leaving it out for now but, like you mentioned, the current coverage isn't great so I want to be mindful of any changes that would potentially make it even worse.

data/stories/chitchat.yml
b-quachtran: In the context of this PR, I think either the `explain` intent should be re-factored to handle the scope of `nlu_info` or the safety net pattern should be left out for now to avoid confusion from a best practices perspective.

samsucik: Thanks @b-quachtran, makes sense. I guess the issue is that we think/care about two different things.
- I think primarily about the new Sara test set, which isn't supposed to be perfect, will inevitably become out-dated soon, and should be used with a particular "archived" snapshot of `rasa-demo` (for model regression purposes).
- You care about `rasa-demo`, which is an evolving thing.

Personally, for the purposes of the new Sara test set, I would just stop here, perhaps reverting the "safety net" changes. But for `rasa-demo`, I totally agree with merging `nlu_info` and `explain` (and altering the `nlu_part` entity). I guess what I'll do for now is leaving only those training story changes in this PR, which were inspired by real conversations (i.e. no "safety net" stuff). And then merging `nlu_info` and `explain` can be another issue, which won't be related to the new Sara test set at all. Sounds good?

samsucik:
> Are there any plans to merge some version of the new test set into `rasa-demo`?

Not that I know of. The other thing is: most of the test set stories will become out-dated once Sara's flows change... I guess what I'll do is adding a few stories from the new test set to this PR. For the sake of not making the coverage worse short-term 🙂

data/stories/chitchat.yml
samsucik: Alright, I've reverted the changes that weren't backed by real stories we have at hand today.

b-quachtran (approved):
Nice! Thanks for getting these changes and additional tests in

b-quachtran:
Looks like CI is failing because of Rasa 2.3.4 and a SocketIO bug, will upgrade to 2.5.X before merging

PR #630, opened 2021-04-14
Fix infinite loop in custom two-stage fallback

hsm207:
Fixes the bug where when the user denies the rephrased intent, the bot will keep asking the 
user to rephrase the user's utterance

### DIFF
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -475,8 +475,8 @@ class ActionDefaultFallback(Action):
 
         # Fallback caused by TwoStageFallbackPolicy
         if (
-            len(tracker.events) >= 4
-            and tracker.events[-4].get("name") == "action_default_ask_affirmation"
+            len(tracker.events) >= 5
+            and tracker.events[-5].get("name") == "action_two_stage_fallback"
         ):
 
             dispatcher.utter_message(template="utter_restart_with_button")
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -501,6 +501,19 @@ class ActionDefaultFallback(Action):
             return [UserUtteranceReverted()]
 
 
+class ActionRestartWithButton(Action):
+    def name(self) -> Text:
+        return "action_restart_with_button"
+
+    def run(
+        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: DomainDict,
+    ) -> None:
+
+        dispatcher.utter_message(template="utter_restart_with_button")
+
+        # return [ConversationPaused()]
+
+
 class ActionCommunityEvent(Action):
     """Utters Rasa community events."""
 
--- a/actions/actions.py
+++ b/actions/actions.py
@@ -36,7 +36,10 @@ class ActionSubmitSubscribeNewsletterForm(Action):
         return "action_submit_subscribe_newsletter_form"
 
     def run(
-        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: DomainDict,
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: DomainDict,
     ) -> List[EventType]:
         """Once we have an email, attempt to add it to the database"""
 
@@ -77,7 +80,10 @@ class ActionSubmitSalesForm(Action):
         return "action_submit_sales_form"
 
     def run(
-        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: DomainDict,
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: DomainDict,
     ) -> List[EventType]:
         """Once we have all the information, attempt to add it to the
         Google Drive database"""
@@ -101,7 +107,8 @@ class ActionSubmitSalesForm(Action):
             return []
         except Exception as e:
             logger.error(
-                f"Failed to write data to gdocs. Error: {e.message}", exc_info=True,
+                f"Failed to write data to gdocs. Error: {e.message}",
+                exc_info=True,
             )
             dispatcher.utter_message(template="utter_salesrequest_failed")
             return []
@@ -133,7 +140,10 @@ class ActionExplainSalesForm(Action):
         return "action_explain_sales_form"
 
     def run(
-        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: DomainDict,
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: DomainDict,
     ) -> List[EventType]:
         requested_slot = tracker.get_slot("requested_slot")
 
@@ -158,7 +168,10 @@ class ActionExplainFaqs(Action):
         return "action_explain_faq"
 
     def run(
-        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: DomainDict,
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: DomainDict,
     ) -> List[EventType]:
         topic = tracker.get_slot("faq")
 
@@ -177,7 +190,10 @@ class ActionSetFaqSlot(Action):
         return "action_set_faq_slot"
 
     def run(
-        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: DomainDict,
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: DomainDict,
     ) -> List[EventType]:
         full_intent = (
             tracker.latest_message.get("response_selector", {})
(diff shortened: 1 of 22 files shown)


ArjaanBuijk:
@hsm207 ,

what is the conversation I should test?

I tried this:
```bash
Bot loaded. Type a message and press enter (use '/stop' to exit): 
Your input ->  I want to know more about rasa 
```

But the custom action throws an exception:
```bash
Exception occurred while handling uri: 'http://localhost:5055/webhook'
Traceback (most recent call last):
  File "/home/arjaan/miniconda3/envs/rasa-demo/lib/python3.7/site-packages/sanic/app.py", line 931, in handle_request
    response = await response
  File "/home/arjaan/miniconda3/envs/rasa-demo/lib/python3.7/site-packages/rasa_sdk/endpoint.py", line 103, in webhook
    result = await executor.run(action_call)
  File "/home/arjaan/miniconda3/envs/rasa-demo/lib/python3.7/site-packages/rasa_sdk/executor.py", line 390, in run
    action(dispatcher, tracker, domain)
  File "/home/arjaan/rasa/repos/rasa-demo/actions/actions.py", line 434, in run
    button_title = self.get_button_title(intent, entities)
  File "/home/arjaan/rasa/repos/rasa-demo/actions/actions.py", line 465, in get_button_title
    return button_title.format(**entities)
AttributeError: 'NoneType' object has no attribute 'format'

```

It does this also in the main branch, so it is does not seem related to your update.

hsm207:
@ArjaanBuijk 

You have found another bug with the custom action which I am also looking into 🤣 I found that a workaround to this issue is to start the conversation with "hi".

To test for the bug that this PR is meant to fix, you can start a conversation with "hi" and then utter something that triggers the nlu_fallback intent e.g. "not this time" to start the two stage fallback process.

The bug happens when you try to reach [step 4 ](https://rasa.com/docs/rasa/fallback-handoff/#two-stage-fallback)of the Two Stage Fallback process. In the main branch, if you deny the rephrased intent, the bot will keep asking you to rephrase instead of triggered an ultimate fallback action.

ArjaanBuijk:
The first time I say a sentence it does not respond anything. It just listens again. 
Then, the second time I say it, it triggers the fallback. This happens for both fallback stages, like in this conversation:

```bash
Your input ->  hi                                                                                             
Hi, I'm Sara!
By chatting to me you agree to our [privacy policy](https://rasa.com/privacy-policy/).
If you're new to Rasa, I can help you get started! If you've already started building an assistant, you can also ask me about the different parts of Rasa 🐦
Your input ->  I want to know more about rasa                                                                 
Your input ->  I want to know more about rasa                                                                 
Could you rephrase your message? I didn't get it, I'm sorry.
Your input ->  I want to know more about rasa                                                                 
Your input ->  I want to know more about rasa                                                                 
Sorry, I didn't get that 🤷. Could you please rephrase?
```

hsm207:
Did you reduce the training parameters for the policy pipeline?

It looks like the rephrase message is triggered by the Core  (the[ else part](https://github.com/RasaHQ/rasa-demo/blob/5cfbaa5545cba41d44b4f7893721bf4e6f2ce8e8/actions/actions.py#L487-L489) in the custom `action_default_fallback).

If the utterance had really triggered Two-Stage Fallback, you should have been guiding from [step 1 through 3](https://rasa.com/docs/rasa/fallback-handoff/#two-stage-fallback) first before `action_default_fallback` gets called in step 4 after you deny the rephrased intent.

hsm207:
@ArjaanBuijk I wrote some unit test to clarify the bug fix

ArjaanBuijk (commented):
@hsm207 ,
I was able to verify the two-stage fallback by using this conversation, but at the end, when I select to start over, the bot is not actually restarting:

```bash
Your input ->  hi                                                                                      
Hey there, my name is Sara.
By chatting to me you agree to our [privacy policy](https://rasa.com/privacy-policy/).
If you're new to Rasa, I can help you get started! If you've already started building an assistant, you can also ask me about the different parts of Rasa 🐦
Your input ->  why go there                                                                            
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  3: Something else (/out_of_scope
)                                                                                                      
Can you rephrase the question?
Your input ->  why go there                                                                            
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  3: Something else (/out_of_scope
)                                                                                                      
? Click the button below if you want to start over.  1: Restart (/restart)                             
Your input ->       (NOTE: bot did not restart...)
```

When I manually type `/restart` it restarts properly:
```bash
? Click the button below if you want to start over.  1: Restart (/restart)                             
Your input ->  /restart                                                                                
Loading... And I've been restarted! 🤯
Your input ->                            
```

Do we need another `rule` or `story` to make sure the bot actually restarts?

hsm207:
@ArjaanBuijk 

I think the start over behavior you observed is expected.

The current implementation of starting over is to set the `feedback_value` slot to `negative` and pause the tracker i.e. call the `ConversationPaused()` event. So the bot will ignore any subsequent user utterance.

`/restart` calls the `[Restarted()](https://github.com/RasaHQ/rasa/blob/7625fcaec9a5d1f47dbcc093a405e9369e7261fc/rasa/shared/core/events.py#L970)` event which resets the tracker.

Should we replace `ConversationPaused()` with `Restarted()`?

ArjaanBuijk:
@hsm207 ,
I fixed several more things:
- the intent name extraction for responseselectors was wrong. This lead to crashes in ask-rephrase
- updated all stories & rules that contained the slot `feedback_value`, because that slot no longer influences the conversation

Now the 2 stage fallback seems to work:
```bash
Bot loaded. Type a message and press enter (use '/stop' to exit): 
Your input ->  how do i trigger

#######################################
#  Stage 1: Ask Rephrase is triggered # 
#######################################                                                                                                                                                 
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  (Use arrow keys)                                                                                               
   1: How do you feel? (How do you feel?)
 » 2: Something else (/out_of_scope)
   Type out your own message... 


#################################################
# Stage 2: action_default_fallback is triggered # 
#################################################   
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  2: Something else (/out_of_scope)                                                                              
Can you rephrase the question?
Your input ->  how do i trigger                                                                                                                                                       
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  (Use arrow keys)                                                                                               
   1: How do you feel? (How do you feel?)
 » 2: Something else (/out_of_scope)
   Type out your own message...      

##############################################################################
# active_loop becomes null & 2nd rule predicts the utter_restart_with_button # 
##############################################################################
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  2: Something else (/out_of_scope)                                                                              
Can you rephrase the question?
Your input ->  how do i trigger                                                                                                                                                       
? Sorry, I'm not sure I've understood you correctly 🤔 Do you mean...  2: Something else (/out_of_scope)                                                                              
? Click the button below if you want to start over.  (Use arrow keys)                                                                                                                 
 » 1: Restart (/restart)
   Type out your own message...

################################
# and now it actually restarts # 
################################
? Click the button below if you want to start over.  1: Restart (/restart)                                                                                                            
Congrats you've restarted me! 😉
Your input -> 
```

Still to do now is to make sure that the stories & rules were correctly updated to reflect that the `feedback_value` slot is not longer influencing the conversations. I will run some tests on this.

ArjaanBuijk (approved):
Everything now passes the ci/cd tests.

actions/actions.py
```diff
@@ -414,37 +462,52 @@ def run(
...
-        entities = tracker.latest_message.get("entities", [])
-        entities = {e["entity"]: e["value"] for e in entities}
-
-        entities_json = json.dumps(entities)
+        if "nlu_fallback" in first_intent_names:
+            first_intent_names.remove("nlu_fallback")
+        if "/out_of_scope" in first_intent_names:
+            first_intent_names.remove("/out_of_scope")
+        if "out_of_scope" in first_intent_names:
+            first_intent_names.remove("out_of_scope")
```
hsm207, line 477: how does this action's run() method work?

config.yml
```diff
@@ -5,7 +5,6 @@ pipeline:
 - name: LexicalSyntacticFeaturizer
 - name: CountVectorsFeaturizer
   OOV_token: oov
-  token_pattern: (?u)\b\w+\b
```
hsm207, line 8: Why do we need this change?

ArjaanBuijk: @hsm207 ,
See [this discussion](https://rasa-hq.slack.com/archives/C4Q4X95M4/p1619090669047600?thread_ts=1619089685.047300&cid=C4Q4X95M4):

It's not a big issue at all, only reasons it should to be removed are to avoid confusing users and because it produces a warning during parameter validation

This is actually the pattern applied when the analyzer parameter is set to word, which is the default:
```
token_pattern=r"(?u)\b\w+\b"
if parameters["analyzer"] == "word"
```

 In short, you can just remove it and the behaviour should stay the same.

hsm207: thanks for explaining!

hsm207:
Solved in #637

PR #638, opened 2021-05-03
Fix playground form

melindaloubser1:
Actually activate the playground form when a playground issue is reported.

This change got accidentally reverted in the new-use-case PR.


melindaloubser1:
@ArjaanBuijk do you have time to look at this today/tomorrow?

ArjaanBuijk (approved):
@melindaloubser1 ,
This all looks good.

Is there an e2e test for this flow?

melindaloubser1:
good catch, added one

PR #645, opened 2021-06-23
Set less requests for rasa worker so it gets scheduled on node with memory tag

melindaloubser1:
* Lowers the CPU/memory requests for rasa-worker so that a second pod can be scheduled on the single extra-memory node in the cluster
The node's label ensures rasa-worker still gets scheduled only on a high memory node.

It was often failing to schedule the node since the existing requests would not leave quite enough for a second pod with the same requirements.


b-quachtran (approved):
🚂

PR #647, opened 2021-07-22
bump to 2.8.0

akelad:


.github/workflows/continuous_deployment.yml
```diff
@@ -34,7 +34,7 @@ env:
   # * rasa version in requirements.txt
   # * rasa-sdk base image version in Dockerfile
   RASA_X_VERSION: "0.40.1"
```
melindaloubser1, line 36: We'll need to wait till we can bump this before merging since 2.8.0 is too high for 0.40.1

melindaloubser1, line 36:
```suggestion
  RASA_X_VERSION: "0.42.0"
```

melindaloubser1 (approved):

.github/workflows/continuous_deployment.yml
```diff
@@ -33,8 +33,8 @@ env:
   # NB! The rasa/rasa-x versions should be in sync with:
   # * rasa version in requirements.txt
   # * rasa-sdk base image version in Dockerfile
-  RASA_X_VERSION: "0.40.1"
-  RASA_VERSION: "2.6.3"
+  RASA_X_VERSION: "0.42.0"
+  RASA_VERSION: "2.8.0"
```
melindaloubser1, line 37:
```suggestion
  RASA_VERSION: "2.8.2"
```

Dockerfile
```diff
@@ -2,7 +2,7 @@
 # * rasa version in requirements.txt
 # * RASA_VERSION and RASA_X_VERSION  in .github/workflows/continuous-deployment.yml
 # Pull SDK image as base image
-FROM rasa/rasa-sdk:2.6.0
+FROM rasa/rasa-sdk:2.8.0
```
melindaloubser1, line 5:
```suggestion
FROM rasa/rasa-sdk:2.8.2
```

requirements.txt
```diff
@@ -1,4 +1,4 @@
-rasa==2.6.2 # NB! when updating, make sure to also update:
+rasa==2.8.0 # NB! when updating, make sure to also update:
```
melindaloubser1, line 1:
```suggestion
rasa==2.8.2 # NB! when updating, make sure to also update:
```

melindaloubser1 (approved):

akelad:
@melindaloubser1 should this be merged?

melindaloubser1:
Yes, it should, but it never did because the discourse API calls in the unit tests are failing

we can force merge it and open an issue to look into that separately, since it means the existing version's calls also won't work anyway.

akelad:
I would rather we fix it, if we have the time 😅

melindaloubser1:
Ok, in that case I can get to it later this week

PR #651, opened 2021-08-27
Test both NLU and Core in Test End 2 End Strories

hsm207:
Update CI command for testing stories to include NLU (it was explicitly doing `test core`)


HotThoughts:
Have you tested `rasa test --stories tests/test_conversations.yml --fail-on-prediction-errors` locally?

I am asking this because I tried to do the same thing for [`rasa-train-test-gha`](https://github.com/RasaHQ/rasa-train-test-gha) but `rasa test` could not pick up the right stories when testing core. 

I ended up run `rasa test nlu` and `rasa test core --stories tests/test_conversations.yml --fail-on-prediction-errors` separately. 

 Seems like `rasa test` does not accept `core` and `nlu` argument, although their arguments are two disjoint sets.  👐

PR #654, opened 2021-09-10
assume there are no events if html tag not found

melindaloubser1:
Currently there are no events posted on the events page, the html tag is not found and therefore the action errors out.

This PR hot-fixes it by assuming when the tag is not found, there are no events.


melindaloubser1:
@hsm207 this should allow your PR to merge too once updated

hsm207 (approved):

PR #659, opened 2021-10-29
Use existing PVCs for rasa x and postgres in deployment

melindaloubser1:
Restore data from existing PVCs with `retain` policies


b-quachtran (approved):
🚀

Issue #657, opened 2021-10-12
cd: investigate model uploading issue


HotThoughts:
Investigate what is going on with the model uploading step in the continuous deployment workflow.

So far we have seen two issues:
```
curl: (52) Empty reply from server
```
https://github.com/RasaHQ/rasa-demo/runs/3859020182?check_suite_focus=true

and

```
***"reasons":["Auth required."],"exception":"Unauthorized"***
```
https://github.com/RasaHQ/rasa-demo/runs/3583794295?check_suite_focus=true

PR #663, opened 2021-11-09
cd: use rasactl to operate models

HotThoughts:
Fix https://github.com/RasaHQ/rasa-demo/issues/657

### DIFF
--- a/.github/workflows/continuous_deployment.yml
+++ b/.github/workflows/continuous_deployment.yml
@@ -229,9 +229,11 @@ jobs:
         RASACTL_RASA_X_URL: https://${{ env.RASA_X_DOMAIN }}
         RASACTL_AUTH_USER: ${{ env.RASA_X_USERNAME }}
         RASACTL_AUTH_PASSWORD: ${{ env.RASA_X_PASSWORD }}
+        # Disable http2client. Current Rasa X (2.6) endpoint doesn’t support http2.
+        GODEBUG: http2client=0
       timeout-minutes: 10
       run: |
-        # Disable http2client. Current Rasa X (2.6) endpoint doesn’t support http2.
-        GODEBUG=http2client=0 rasactl model upload ${MODEL_PATH}
+        # Upload model
+        rasactl model upload ${MODEL_PATH}
         # Tag model as production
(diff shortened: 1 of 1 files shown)


HotThoughts:
@tczekajlo tag you here because you know rasactl the best. Have you seen [this error](https://github.com/RasaHQ/rasa-demo/runs/4155251456?check_suite_focus=true) before? 
```console
Error: Post "https://***/api/projects/default/models": http2: server sent GOAWAY and closed the connection; LastStreamID=5, ErrCode=ENHANCE_YOUR_CALM, debug=""
```

Is it about the header field size for http2 configured on nginx server?

HotThoughts:
@tczekajlo @melindaloubser1 

💫 Ready for review. See [this](https://github.com/RasaHQ/rasa-demo/runs/4225880413?check_suite_focus=true) successful run. I am so happy that `rasactl` worked here. Love it.

.github/workflows/continuous_deployment.yml
```diff
@@ -146,6 +146,7 @@ jobs:
 
     - name: Prepare namespace
       run: |
+        kubectl label ns "${NAMESPACE}" deployment=sara || true
```
melindaloubser1, line 149: What is this for?

melindaloubser1 (approved):
Just one Q about what the ns line is for, otherwise glad it works !

.github/workflows/continuous_deployment.yml
HotThoughts: This label is added to ensure that the namespace has at least one label, which helps `rasactl` to request namespace information properly. 

`kubectl` won't give you the requested json data if the namespace does not have any label. 🤷🏻‍♀️

tczekajlo: It was a workaround, a bug was fixed in the latest version of rasactl. @HotThoughts we can remove this line, it's not needed anymore.

tczekajlo:
```suggestion
```

HotThoughts: Wow. That was fast. Thank you @tczekajlo for fixing it. I will remote this line.

PR #668, opened 2022-01-26
Update to 1.0

melindaloubser1:
Update to Rasa XE 1.0.1
Database has already been manually migrated to Postgres 12.9

### DIFF
--- a/.github/deployments/helmfile.yaml
+++ b/.github/deployments/helmfile.yaml
@@ -59,6 +59,10 @@ releases:
                 name: {{ requiredEnv "ACTION_SERVER_SECRET_NAME" }}
                 key: "ALGOLIA_DOCS_INDEX"
       - postgresql:
+          tag: "12.9.0"
+          extraEnv:
+            - name: PGDATA
+              value: /bitnami/postgresql/data_12
           persistence:
             existingClaim: {{ requiredEnv "POSTGRES_PVC" }}
       - global:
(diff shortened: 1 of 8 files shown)


.github/deployments/helmfile.yaml
```diff
@@ -59,6 +59,10 @@ releases:
                 name: {{ requiredEnv "ACTION_SERVER_SECRET_NAME" }}
                 key: "ALGOLIA_DOCS_INDEX"
       - postgresql:
+          tag: "12.9.0"
+          extraEnv:
+            - name: PGDATA
```
hsm207, line 64: Why do we need this env var now?

.github/deployments/helmfile.yaml
```diff
@@ -111,8 +116,9 @@ releases:
               session_persistence: true
             rest: {}
       - rabbitmq:
-          rabbitmq:
+          auth:
             password: {{ requiredEnv "RABBITMQ_PASSWORD" }}
+            erlangCookie: {{ requiredEnv "RABBITMQ_ERLANGCOOKIE" }}
```
hsm207, line 121: Why do we need this env var now?

melindaloubser1: The updated Rabbit chart requires it

.github/deployments/helmfile.yaml
melindaloubser1: For the manual db migration to 12.9 it was required; I tried removing it subsequently but that makes it error out

hsm207 (approved):

hsm207 (approved):

PR #732, opened 2023-12-15
python3 command

johnycage:


johnycage:
python3 command is needed for latest version. Suggested changes in Makefile.

trammell:
I can't think of any OS or environment that uses python 2.x these days. Is this really a problem?

johnycage:
> I can't think of any OS or environment that uses python 2.x these days. Is this really a problem?

I don't have 'python-is-python3' package installed on my Ubuntu 22.04 and so for any default installation on Ubuntu python3 is goto command. (python i.e. python2 is not installed).
May be a note can be added, so make install can run smoothly.

trammell:
One possible workaround for you would be to create a shell alias, something like

```bash
alias python=/usr/bin/python3.10
```

johnycage:
> One possible workaround for you would be to create a shell alias, something like
> 
> ```shell
> alias python=/usr/bin/python3.10
> ```

That's right, but then we must mention in the setup documentation about python3 caveat. Let's add a line in readme maybe.

